PR #1, opened 2019-10-13
Project - 5

natalieschnelzer:
Changes in javascript, css and html stay the same

### DIFF
--- a/code/clue.js
+++ b/code/clue.js
@@ -1,25 +1,93 @@
-// CREATE OBJECTS FOR ALL THE SUSPECTS, SOMETHING LIKE THIS:
+// SUSPECTS AS OBJECTS
 
 const mrGreen = {
-  firstName: "Jacob",
+  title: "Mr",
   lastName: "Green",
   color: "green",
-  description: "He has a lot of connections",
+  description: "Hypocritical Anglican priest",
   age: 45,
   image: "assets/green.png",
   occupation: "Entrepreneur"
 }
 
-// CREATE OBJECTS FOR ALL THE WEAPONS, ADD MORE CHARACTERISTICS THE THE WEAPONS IF YOU LIKE.
+const mrsWhite = {
+  title: "Mrs",
+  lastName: "White",
+  color: "white",
+  description: "Frazzled",
+  age: 54,
+  image: "assets/white.png",
+  occupation: "Servant"
+}
 
-const rope = {
-  name: "rope",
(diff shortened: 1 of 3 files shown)


code/clue.js
```diff
@@ -1,64 +1,179 @@
...
+  'study',
+  'hall',
+  'lounge',
+  'dining room'
 ]
 
-const weapons = []
+let windowKiller = null 
+let windowWeapon = null
+let windowRoom = null
```
JennieDalgren, line 123: how come you removed the mystery object?
This is fine, I'm just curious.

code/clue.js
```diff
@@ -1,64 +1,179 @@
...
 
-const rooms = []
 
-// THIS FUNCTION WILL RANDOMLY SELECT ONE ITEM FOR THE ARRAY THAT YOU PASS IN TO THE FUNCTION.
-const randomSelector = array => {
-  return array[Math.floor(Math.random() * array.length)]
+// SELECTING RANDOM 
+function randomOptionSelector (optionsArray) {
+  const randomIndex = Math.floor(Math.random() * optionsArray.length)
+  return optionsArray[randomIndex]
```
JennieDalgren, line 129: how come you moved away from es6 arrow function syntax?

code/clue.js
```diff
@@ -1,64 +1,179 @@
...
 }
 
-// CREATE OBJECTS FOR ALL THE WEAPONS, ADD MORE CHARACTERISTICS THE THE WEAPONS IF YOU LIKE.
+const mrsWhite = {
+  title: "Mrs",
+  lastName: "White",
+  color: "white",
+  description: "Frazzled",
+  age: 54,
+  image: "assets/white.png",
```
JennieDalgren, line 19: would be cool to see if you can get the images to show on the cards as well? 😎

JennieDalgren (approved):

code/clue.js
natalieschnelzer: Tiago was helping me understand the flow so we changed the names and broke it down in smaller pieces.

code/clue.js
natalieschnelzer: Also, a way to understand and break it down a little bit further. Will be going through the code again and double-check these things and try to understand them for es6

code/clue.js
natalieschnelzer: Yes!! It's on my to-do list!

PR #5, opened 2019-10-24
Project 5 Angelica

Angelicabrodin:
I have decided to hand in my project without going deeper into the stretch goals for now. 

I found it very tricky to understand the differences between arrays and objects and when it is a string and not and when you therefore need "" or not. But I think I got the hang of it for this assignment. 

I also struggled alot with the mystery object and how it was supposed to look, I found it very confusing that you actually did not add any content but that it was instead stored there when invoking the different cards or decks. 

And finally, I really struggled with invoking the extra information about the killer and the weapon, for some reason I needed to change from "" to ' ' and I still think that it is also very confusing :) so I experimentet alot before I got the help and were able to change it so it actually came up in the browser. 

All in all, it was a very fun and developing assignment, even though it is still hard to grasp everything :)


Angelicabrodin:
This is my site on netlify: 
angiescluedogame.netlify.com

code/clue.js
```diff
@@ -5,36 +5,159 @@ const mrGreen = {
...
+  candlestick,
+  dumbbell,
+  poison,
+  axe,
+  bat,
+  trophy,
+  pistol,
+];
+
+//These are strings not objects, therefore " "
```
ericaglimsholt, line 139: Is this comment for the `rooms` array? Then they are strings and not objects. So you have wrote the array correctly!

code/clue.js
```diff
@@ -43,22 +166,57 @@ const randomSelector = array => {
 // With a killer, a weapon and a room.
 // The values will be set later.
 
+const mystery = {
+};
```
ericaglimsholt, line 170: Are you using this object somewhere?

ericaglimsholt (approved):
Hi Angelica! Creat work on the assignment 🚀
I understand that it can be confusing with all of this, it's still pretty new to you guys. About the issue with `''` and `""` - even I haven't really get the hang of it. I understand it 8/10 times, but sometimes it makes no sense. A good thing to remember is that you don't need to know everything by hand, but if you know for an example that there're a difference between `''` and `""` and your code doesn't work - try change them out for the other one. Then you know if that's the reason why the code doesn't work. This part, as well as the rest, is something that you will get a hang on and learning by doing is the way to go when it's about programming.

code/clue.js
Angelicabrodin: I wrote the comment for myself, to understand what I was doing, and luckily they were right then :)

code/clue.js
Angelicabrodin: This is the object that all get stored in, before that data or input gets printed on the screen in the h1 element when pushing the button, have I put it in the wrong order maybe?

ericaglimsholt: Okay! No it's fine just wanted to know how you use it! Cool 😁

PR #6, opened 2019-10-24
Project 5 - Fredrik Tibbling

codebyfredrik:
Project on Netlify: https://hardcore-mcnulty-bcfdea.netlify.com/


ericaglimsholt (approved):
Great work Fredrik! 🚀
Really enjoyed trying out your game. And a plus for getting some nice css work done to this assignment. If you want to develop the page even more I would get into making the page more responsive. Of course it's responsive, I can see your use of `@media`, but what I mean is that you can create a better performance for the user in the mobile. Now the user has to scroll a lot to run this game. So maybe there's a way where the user on a mobile can play your game a little more smoother?

codebyfredrik:
Thanks for great feedback, Erica! 
I totally agree with you that I could have been a little bit more creative from a design standpoint going from desktop, tablet to mobile. It's way to easy to just change flex-direction from row to column and tick the box for responsive design. I do realize the game could be more user friendly on mobile. Coming up with great design ideas is sometimes hard. I try to challenge myself in each project to push myself forward. I use Pinterest and other sources to come up with design ideas. It's a skill that I will hopefully be able to develop over time.

PR #7, opened 2019-10-25
Cluedo Emma Berglund

EmmaBerglund1977:


ericaglimsholt (changes requested):
Hi Emma! Can you include your Netlify link in the PR description?

EmmaBerglund1977:
https://fervent-davinci-548aa4.netlify.com/

code/clue.js
```diff
@@ -43,22 +175,63 @@ const randomSelector = array => {
...
 // 'The murder was committed by Jacob Green, in the living room with a rope.'
+
+const revealMystery = () => {
+  const myst = document.getElementById('mystery')
+  myst.innerHTML = `The murder was committed by ${mystery.killer.firstName +
+    ' ' +
+    mystery.killer.lastName}, in the ${mystery.room} with a ${
+    mystery.weapon.name
+  }`
+}
```
ericaglimsholt, line 237: When pushing the button (running this function) without clicking on the cards you get back `The murder was committed by undefined undefined, in the [object HTMLDivElement] with a undefined` as a string. And that's maybe not the best. What you can to is to check if `mystery.killer.firstName`, `mystery.killer.lastName`, `mystery.room` and `mystery.weapon.name` has any value. If they do - output the string. If they don't do nothing or add an error message for the user

code/clue.js
```diff
@@ -1,40 +1,172 @@
...
 
 const rope = {
-  name: "rope",
-  weight: 10
+  name: 'rope',
+  weight: 500
+}
+
+const knife = {
+  name: ' knife',
```
ericaglimsholt, line 70:
```suggestion
  name: 'knife',
```

code/clue.js
```diff
@@ -1,40 +1,172 @@
...
+const billiardRoom = 'Billiard Room'
+const lounge = 'Lounge'
+const ballRoom = 'Ball room'
+const hall = 'Hall'
+const spa = 'Spa'
+const livingRoom = 'Living Room'
+const observatory = 'Observatory'
+const theater = 'Theater'
+const guestHouse = 'Guest House'
+const patio = 'Patio'
```
ericaglimsholt, line 125: Because you only have this as simple strings, you can add it directly to your array instead of declaring every room as it's own property

code/clue.js
```diff
@@ -43,22 +175,63 @@ const randomSelector = array => {
...
+  const theChosenWeaponName = document.getElementById('weaponName')
+
+  theWeapon.style.background = mystery.weapon.color
+  theChosenWeaponName.innerHTML = mystery.weapon.name
+}
 
+// FUNCTION pickRoom
+
+const pickRoom = () => {
+  //
```
ericaglimsholt, line 220: Empty comment? :)

ericaglimsholt (approved):
Great work! Added some comments for you to think about :)

PR #8, opened 2019-10-25
Project 5 Cluedo

nyblad:
Live site: https://project-5-cluedo.netlify.com/


code/clue.js
```diff
@@ -1,26 +1,131 @@
...
+// OBJECTS FOR ALL THE SUSPECTS
 const mrGreen = {
   firstName: "Jacob",
   lastName: "Green",
-  color: "green",
+  color: "rgba(48, 75, 59, 0.8)",
   description: "He has a lot of connections",
-  age: 45,
-  image: "assets/green.png",
+  age: "Age: 45",
```
ericaglimsholt, line 7: As of age being a number you can add this value as a int instead of a string. The difference is that a string is for example a word and needs `''` or `""` around it. An int doesn't need `''` or `""` around it - it can be clear numbers. So in this example you can type it as `age: 45`

code/clue.js
```diff
@@ -1,26 +1,131 @@
...
+  description: "She is a frazzled servant",
+  age: "Age: 65",
+  image: "assets/white.jpg",
+  occupation: "Housekeeper"
+}
+
+// OBJECTS FOR ALL THE WEAPONS
+const rope = {
+  name: "Rope",
+  weight: "Weight: 2kg",
```
ericaglimsholt, line 65: The same goes here as before on the `age`. It seems like you have decided that the weight always is in `kg`, so you can use just numbers here as well. So this could be `weight: 2`

code/clue.js
```diff
@@ -30,9 +135,38 @@ const suspects = [
...
+  "Spa",
+  "Living Room",
+  "Observatory",
+  "Theater",
+  "Guest House",
+  "Patio"
+]
+
+//Console.log to check the arrays
+console.log(suspects, weapons, rooms)
```
ericaglimsholt, line 169: Remember to delete `console.logs` before submitting it in a PR!

ericaglimsholt (approved):
Great work on your Cluedo assignment! 🚀
One thing that you can take a better look at it how use string and int. I've giving you a small explanation below if that helps you in any way. But wow, you have really created a well looking page! It's so fun to see that you have been creative about this. The styling looks fab!

PR #9, opened 2019-10-25
Cluedo Project 

Agnes94:
My indentation might be weird at some places because of Prettier. 
https://thirsty-swanson-5f9579.netlify.com/


code/clue.js
```diff
@@ -8,57 +8,225 @@ const mrGreen = {
...
+  "Observatory",
+  "Theater",
+  "GuestHouse",
+  "Patio"
+];
 
-const rooms = []
+console.log(suspects);
+console.log(weapons);
+console.log(rooms);
```
ericaglimsholt, line 155: Make sure to delete the `console.logs` before submitting a PR!

ericaglimsholt (approved):
God job! 🚀
I think you did a great work. I like that you have styled the cards in a different way and that you have included the background image. I think you should keep up the good job. I can't really see that you have been struggling with this. If you want to make it even better you could work on the responsiveness of the page - now it doesn't work on mobiles.

PR #13, opened 2019-10-26
Clara Gyökeres Cluedo

claragyokeres:
Here's my pull request!


ericaglimsholt (changes requested):
Please add a Netlify link to your PR!

claragyokeres:
https://clarascluedo.netlify.com

code/clue.js
```diff
@@ -43,22 +157,71 @@ const randomSelector = array => {
...
+
 // CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
 // 'The murder was committed by Jacob Green, in the living room with a rope.'
+
+const revealMystery = () => {
+
+  const theMystery = document.getElementById("mystery")
+
+  theMystery.innerHTML =
+    (`The murder was committed by ${mystery.killer.firstName} ${mystery.killer.lastName}, in the ${mystery.room} with a ${mystery.weapon.name}.`)
```
ericaglimsholt, line 225: Here you got an error when you are clicking on the button, but hasn't chosen a room, killer and weapon. For handling this you can do an if statement here. For example you can check if `mystery.killer.firstName` exists

code/clue.js
```diff
@@ -43,22 +157,71 @@ const randomSelector = array => {
...
+  mystery.room = randomSelector(rooms)
+
+  const theRoom = document.getElementById("roomName")
+
+  theRoom.innerHTML =
+    mystery.room
+
+}
+
+console.log(pickRoom);
```
ericaglimsholt, line 215:
```suggestion
console.log(pickRoom());
```

ericaglimsholt (approved):
Hey Clara! Great work with your Cluedo assignment. What you can try to take with you for the next assignment is error handling. So if something doesn't work like it suppost to we need to handle that in some way. Usually we say to the user that this behavoir is wrong and a description how the user should do it. In this example an `alert` saying `you need to choose a room, killer and weapon to generate a murder` would do

PR #16, opened 2019-10-27
Project 5: Cluedo, Michel Elmoznino Laufer

MichelLaufer:
Fun project!


ericaglimsholt (changes requested):
Please add your Netlify link to the PR!

MichelLaufer:
Hi!

I added the Netlify-link in Canvas, and it is:
https://michel-cluedo.netlify.com

Best,
Michel
________________________________

code/clue.js
```diff
@@ -10,16 +10,120 @@ const mrGreen = {
...
+const conservatory = "Conservatory"; 
+const kitchen = "Kitchen";
+const study = "Study";
+const library = "Library";
+const billiardroom = "Billiard Room";
+const lounge = "Lounge";
+const ballroom = "Ballroom";
+const hall = "Hall";
+const spa = "Spa";
+const livingroom = "Living Room";
```
ericaglimsholt, line 123: If you are only  this as simple strings you can add it directly to your `rooms` variable. This means less code, which you always should aim for. Ex: 
```
const rooms = [
  "Dinning Room",
  "Conservatory",
];
```

code/clue.js
```diff
@@ -52,13 +186,70 @@ const pickKiller = () => {
...
+const evening = "evening";
+const night = "night";
+
+// Creating an array of when the murder was committed
+const times = [
+  morning,
+  afternoon,
+  evening,
+  night
+]
```
ericaglimsholt, line 232: Same goes here as above

ericaglimsholt (approved):
Great work! 🚀 What you can think about is how to make the code less. That's something you always should aim for - less code makes it easier to read and won't take up so much space (because code weights)

PR #24, opened 2019-10-27
Nina's Cluedo

ninamansster:
Here is the published version:
https://ninamansster-cluedo.netlify.com/

Enjoy!

//nina


code/clue.js
```diff
@@ -30,35 +114,132 @@ const suspects = [
...
+  const theKillerImage = document.getElementById("killerImage")
+  const theKillerDescripton = document.getElementById("killerDescription")
 
   theKiller.style.background = mystery.killer.color
   theKillerName.innerHTML =
     mystery.killer.firstName + " " + mystery.killer.lastName
+  theKillerAge.innerHTML = mystery.killer.age
+  theKillerOccupation.innerHTML = mystery.killer.occupation
+  theKillerImage.src = mystery.killer.image
+  theKillerDescripton = ""
```
ericaglimsholt: Here you get a type error! Why? It's because you have declared the `theKillerDescription` variabel with a `const`. Here you can read about what it means: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Invalid_const_assignment

ericaglimsholt (approved):
Great work! You got a type error, which I have comment about and added a link that should be very helpful to you, for knowing what the error means and how to solve it. But otherwise then that, everything looks great!

PR #31, opened 2019-10-28
Project 5 - cluedo game - Jola Chabik

ghost:
https://ucha3i-project-5.netlify.com/

I have a little bit of trouble styling the cards, unfortunately I finish one of them, but I'd like another two works and looks exactly the same.


ericaglimsholt (approved):
I think you did assignment really good! I don't have a lot to comment on. Is there any part that you struggle with? I think that's a good idea to make the other cards in the same way as the one you like, just like you said.

ghost:
I guess my the main struggle was to really understand what I've done, like to feel that I'll be able to reuse the knowledge, not just find an answer on the internet. 

Also, transitions in CSS are still giving me the hard times but we're getting to know it other better ;)

ericaglimsholt:
> I guess my the main struggle was to really understand what I've done, like to feel that I'll be able to reuse the knowledge, not just find an answer on the internet.
> 
> Also, transitions in CSS are still giving me the hard times but we're getting to know it other better ;)

I understand the struggle. For me I don't really  **think** I understand something until I've explained it to someone and answer questions about it. Have you tried that? Ex go through what you have done with a friends who's never seen code before

PR #35, opened 2019-10-28
Project Cluedo

JoacimNilsen:
I used the already made layout and tweaked it a bit in ways of styling, adding background image and some hover effects.
It was fun to start using JS for real and I am happy with the result. I finally understand how it works and now i just want to learn to actually write JS with the same confidant as with CSS and HTML.


ericaglimsholt (changes requested):
Please add the Netlify link to the PR :)

JoacimNilsen:
> Please add the Netlify link to the PR :)

I added it to the readme if that is what you meant? 😬

That's fine, thank you! Next time you can add it to your PR (where you have written the other message on top of this page) :)

code/clue.js
```diff
@@ -28,37 +138,109 @@ const suspects = [
...
+    }, ${mystery.killer.firstName +
+      " " +
+      mystery.killer.lastName} who did it! Using a ${
+      mystery.weapon.name
+    } wich is a ${mystery.weapon.type} weapon and weighs ${
+      mystery.weapon.weight
+    }kg. The gruesome murder took place in the ${mystery.room} `;
+};
+
+document.getElementById("mystery").addEventListener("click");
```
ericaglimsholt, line 246: You get this error when running the project: `Uncaught TypeError: Failed to execute 'addEventListener' on 'EventTarget': 2 arguments required, but only 1 present.`

In this case, this row of code isn't needed what I can see. What was suppost to happen here?

ericaglimsholt (approved):
Nice submission! When you run the project you get an error in your console. The project still works, but we don't want errors on the page, so please make sure the projects doesn't include errors next time :)

PR #37, opened 2019-10-29
Cluedo_Eva-Lisa

microlinamedia:
There are still a few issues to handle:
CSS:
- the header and white bord isn't going straight out to the right edge.
- same with footer but left margin.
- no space between the cards in mobile version, and the cards are squeezed.

JS:
- when clickin on the reveal-button without chosing a card the message "The murder was committed by undefined undefined, in the [object HTMLDivElement] with a undefined." shows


microlinamedia:
https://wizardly-swirles-3c64dd.netlify.com/

idlefingers (approved):

PR #50, opened 2020-02-10
Johanna Åstrand - Project Cluedo

jastrand:
This was a super fun project where you could use skills from previous weeks as well as some new javascript. If I would have more time on the project, I would definitely improve the cards and background image on smaller screens as it's semi-responsive right now and also add some cool flip effects to the cards.

Things I struggle with:
I had some issues with the path of the images that I used on the Cards and Background, finally I had to move them to the same folder as the code to make it to work. I tried everything including /code/assets/path-to-image etc. 

I would also like to learn how to style multiple things at the same time in CSS such as the hover effect I have on the card "decks", I added them to every single one but I am sure there is a smarter way.


code/clue.js
```diff
@@ -4,39 +4,185 @@ const mrGreen = {
...
+const pistol = {
+  name: 'pistol',
+  weight: 30,
+  image: 'assets/weapons/pistol.jpeg'
 }
 
 // THE ROOMS ONLY HAS A NAME SO NO NEED FOR OBJECTS THERE.
 
+/*console.log(suspects.firstname)
+console.log(mrsWhite.description)*/
```
ericaglimsholt, line 123: We can delete code that we doesn't use :)

code/clue.js
```diff
@@ -46,14 +192,72 @@ const pickKiller = () => {
...
+    'weaponImage');
+  theWeaponImage.src = mystery.weapon.image;
+}
+
+//FUNCTION FOR pickRoom:
+
+
+const pickRoom = () => {
+  mystery.room = randomSelector(rooms)
+  document.getElementById('roomCard').style.background = 'darkslategray';
```
ericaglimsholt, line 235: Interesting color! Never heard of `darkslategray ` before! :D

code/clue.js
```diff
@@ -46,14 +192,72 @@ const pickKiller = () => {
...
+  document.getElementById(
+    'timeName'
+  ).innerHTML = `${mystery.time}`
+}
+
 // STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
 // 'The murder was committed by Jacob Green, in the living room with a rope.'
+
+const revealMystery = () => {
+  if (!mystery.killer || !mystery.room || !mystery.weapon || !mystery.time) {
```
ericaglimsholt, line 255: Nice work on the error handling!

code/index.html
```diff
@@ -1,37 +1,51 @@
...
+      <p id="killerAge"></p>
+      <p id="killerDescription"></p>
+      <p id="killerOccupation"></p>
+      <img id="killerImage" src="" height="100" />
+    </article>
+    <article class="weapon-deck" id="weaponCard" onclick="pickWeapon()">
+      <p>WEAPON</p>
+      <p id="weaponName"></p>
+      <p id="weaponWeight"></p>
+      <img id="weaponImage" src="" height="100" />
```
ericaglimsholt, line 29: Try to sett the height in the css instead

ericaglimsholt (approved):
Great work on your cluedo project! It's really nice to see what you have developed. I think your javascript is really good and I'm hoping that you are proud of it!

About the css. I understand that you want to reuse some of the styling - and you should aim for that. In your case you could go for two solutions;

1 ) You can add a second class to your element. Instead of just using `killer-deck` ex you can have `killer-deck deck`, where you set the general deck styling to `.deck`. In that way you can add `deck` to those parts that wants that styling

2 ) In css you can choose multiple classes like this
```
.killer-deck, .weapon-deck {
  // styling
}
```
And for hover it would look like this 
```
.killer-deck:hover, .weapon-deck:hover {
  // styling
}
```

code/clue.js
jastrand: It's pretty cool, I found it here among with other cool ones: https://htmlcolorcodes.com/color-names/

ericaglimsholt: Thank you for sharing!

PR #53, opened 2020-02-11
Project Week 5 - Tuliany

Tuliany:
In the beginning I struggled on understand how to call out functions, but after working with the videos I managed to understand. 
I've come to conclusion that I need to read more about the different ways to call out html elements in javascript.


ericaglimsholt (changes requested):
Hi Tuliany! When testing out your project through your [Netlify link](https://confident-gates-254712.netlify.com/), I can't make it to work. Whatever I choose it changed once I reveal the crime. Therefor I'm adding project as "request changes" (it means that it isn't approved). What I want you to do is update the code so you can reveal the crime with the cards that you have choosen. I will get a notice once you have updated this project again and will see if it works then and approve this project. Let us know if you need guidance to make this work! :)

Tuliany:
I'm fully aware of that, and it's suposed to do that. Or was the requirement that it should change?

ericaglimsholt:
> I'm fully aware of that, and it's suposed to do that. Or was the requirement that it should change?

Hi, yes you need to change that since the project isn't working. The cards that are being showed are not the once that are being used. If you don't understand please write to me in Slack and we can have a chat :)

PR #73, opened 2020-02-13
Cluedo Mystery picker - Jennifer Thorsen

Jenniferot:
I learned a lot from this project! It was really tricky for me and I struggled a lot with step 3 and 4, but when I finally made it I got to understand it better. I particularly learned more about functions and how to connect to the DOM. 

This is the first time I reached some stretch goals! I added images and then styled the page and made it responsive.

What I would like to work more on is how to tidy up my CSS. I would like some ideas of how to do that in the best way.

Link to site: https://lucid-neumann-499ec2.netlify.com/


code/clue.js
```diff
@@ -3,57 +3,287 @@
...
+  // This will randomly select a room from the rooms. And add that to the mystery object.
+  mystery.room = randomSelector(rooms);
+  document.getElementById("roomName").innerHTML = `${mystery.room.name}`
+  document.getElementById('roomImage').src = mystery.room.image;
+  document.getElementById('roomCard').style.background = mystery.room.color;
+}
+
+document.getElementById('killerCard').addEventListener('click', pickKiller)
+document.getElementById('weaponCard').addEventListener('click', pickWeapon)
+document.getElementById('roomCard').addEventListener('click', pickRoom)
```
puj, line 273: Good job using addEventListener!

code/clue.js
```diff
@@ -3,57 +3,287 @@
...
+
 // STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
 // 'The murder was committed by Jacob Green, in the living room with a rope.'
+
+const theMystery = document.getElementById('mystery');
+
+const revealMystery = () => {
+  if (mystery.killer === " " || mystery.weapon === " " || mystery.room === " ") {
```
puj, line 281: Good job!  Using || here saved you some lines of code!

code/index.html
```diff
@@ -1,37 +1,47 @@
...
-    </section>
 
-    <h1 id="mystery"></h1>
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+  <link rel="stylesheet" href="style.css" />
+  <link href="https://fonts.googleapis.com/css?family=Abril+Fatface&display=swap" rel="stylesheet">
+  <link href="https://fonts.googleapis.com/css?family=Bangers&display=swap" rel="stylesheet">
```
puj, line 10: OoOoOoOooo! Google fonts! 👍

code/index.html
```diff
@@ -1,37 +1,47 @@
...
-    <script src="clue.js"></script>
-  </body>
-</html>
+  <section class="cards">
+    <article class="killer-deck" id="killerCard">
+      <h2>- KILLER -</h2>
+      <p id="killerName"></p>
+      <p id="killerAge"></p>
+      <p id="killerOccupation"></p>
+      <img id="killerImage" class="killerImage" src="" height="240" />
```
puj, line 25: Can definitely include the height in CSS if you want :)

code/style.css
```diff
@@ -1,58 +1,264 @@
 body {
   margin: 0;
   font-family: "Courier New", Courier, monospace;
```
puj, line 3: Maybe you can set this to the Font you used the most and change it in specific places where you use another font?

code/style.css
```diff
@@ -1,58 +1,264 @@
...
+  background: rgb(121, 3, 3);
 }
 
-p {
-  margin: 1px;
+h1 {
+  text-align: center;
+  font-family: 'Abril Fatface', cursive;
```
puj, line 9: If you include this in the body, it's unnecessary here

code/style.css
```diff
@@ -1,58 +1,264 @@
...
 }
 
 .weapon-deck {
   width: 250px;
   height: 400px;
-  background: rgb(10, 53, 10);
+  background: rgb(18, 70, 28);
   text-align: center;
-  padding-top: 30px;
+  padding-top: 27px;
```
puj, line 115: If you created another class called `deck` You could include common CSS styles from all the decks

```css
.deck {
  width: 250px;
  height: 400px;
  border-radius: 16px; 
  text-align: center; 
  padding-top: 27px; 
}
```

Then all you need is to change the styles which are specific to each deck.  You would have two classes on the deck elements in your DOM if you do this. 👍

code/style.css
```diff
@@ -1,58 +1,264 @@
...
+    width: 250px;
+    height: 350px;
+    margin: 10px;
+  }
+
+  .room-deck {
+    width: 250px;
+    height: 350px;
+    margin: 10px;
+  }
```
puj, line 188: Great job with the responsiveness.  Oftentimes it's very difficult to reduce the amount of CSS when designing for responsiveness.

puj (approved):
You did a wonderful job styling this! Images for every card and well thought out design.  I liked it.

If you want to practice reducing the amount of repeated CSS, you can read a bit about inheritance. 
https://developer.mozilla.org/en-US/docs/Web/CSS/inheritance
Check out 'Inherited Properties'

The basic concept is that you can set a property once (so it applies to more than one class and/or element type) and only change it if you have an exception.  

Another thing is making an element have multiple classes. 
A CSS class `deck` could have styling attributes common to all decks and `weapon-deck`, `suspects-deck` etc would only have styling which needs to be different for that specific deck.

PR #78, opened 2020-02-14
Project Cluedo by Johanna O

olsenjohanna:
A hard week combined with a limited time to work with this made it pretty hard. I have had a lot going on at work but in the end I manage to put this project together. I had a session today with Van that made very good explanation to things, it is hard do understand how it all fits together and works in the end but todays session made it a lot clearer. If I would have had more time I would have practice more understanding of javascript and also put more effort in design and making the site responsive.


code/clue.js
```diff
@@ -10,50 +10,223 @@ const mrGreen = {
...
+  'observatory',
+  'theater',
+  'guest house',
+  'patio'
+]
+
+// Function for reload of page, to play again
+const reload = () => {
+  location.reload();
+}
```
puj, line 158: Nice, this is good!

code/clue.js
```diff
@@ -10,50 +10,223 @@ const mrGreen = {
...
-// CREATE AN OBJECT THAT KEEPS THE MYSTERY.
+
+// OBJECT THAT KEEPS THE MYSTERY.
 // With a killer, a weapon and a room.
 // The values will be set later.
+const mystery = {
+  killer: 'unknown',
+  weapon: '',
+  room: ''
+}
```
puj, line 176: It would be good to be consistent here.  Assign all of these values the same thing.
Good options would be `null` or `undefined`

```javascript
const mystery = {
  killer: undefined,
  weapon: undefined,
  room: undefined
}
```

This would allow you to make a smarter reveal function on line 227.  To make sure you reveal only after the user has clicked all the cards 👍

code/style.css
```diff
@@ -17,6 +33,13 @@ p {
   display: flex;
   justify-content: space-around;
   align-items: center;
+  background-image: url(assets/hauntedhouse.jpg);
```
puj, line 36: `    background-size: cover;`
might be super useful here.  Give it a try and google what it does 👍

code/style.css
```diff
@@ -37,16 +69,31 @@ p {
   border-radius: 16px;
 }
 
+.weapon-deck:hover {
+  cursor: pointer;
+  opacity: 0.8;
+}
+
```
puj, line 76: Is there any way you could do this for all decks with a single class?  Maybe something to discuss with your team?

puj (approved):
Good job Johanna! I'm glad we talked in the 1:1!  Looking forward to next project already! 😄

PR #79, opened 2020-02-14
Cluedo By Lisa 

L-Lisa:
I spent way too much time being stuck in details in this project, on the bright side I have the feeling loops and arrays will be something I meet again. I enjoyed the format of this project and hope to return with more time to style it. 
I'm still learning about the js code so really appreciate the feedback on how to do it better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Cluedo - Mystery Randomizer
 This is my second week into learning js, this project was supposed to teach me more about loors and arrays. 
-I tried to fucus on learning more than the design, will make tome to come back for that. Boot-camp life is stressful! 
+I tried to fucus on learning more than the design, will make tome to come back for that. Boot-camp life is busy! 
 
 
 ## The problem
--- a/README.md
+++ b/README.md
@@ -1,13 +1,10 @@
 # Cluedo - Mystery Randomizer
+This is my second week into learning js, this project was supposed to teach me more about loors and arrays. 
+I tried to fucus on learning more than the design, will make tome to come back for that. Boot-camp life is stressful! 
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
 
 ## The problem
-
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+I get stuck in the details. I need to be stricter with my time allocation and ask for help when i need it. When I do, I need to practice formulating my questions in the best way.
 
 ## View it live
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
--- a/README.md
+++ b/README.md
@@ -8,3 +8,4 @@ I get stuck in the details. I need to be stricter with my time allocation and as
 
 ## View it live
 
+https://clever-carson-d0f8a3.netlify.com
--- a/code/clue.js
+++ b/code/clue.js
@@ -1,4 +1,5 @@
 // STEP 1 - CREATE OBJECTS FOR ALL THE SUSPECTS, SOMETHING LIKE THIS:
+//
 
 const MrGreen = {
   firstName : 'Jacob',
@@ -32,7 +33,7 @@ const MissScarlet ={
 
 const MrsPeacock = {
   firstName : 'Jane',
-  lastName: 'Doh',
+  lastName: 'Aka - Mrs Peacock',
   color: 'blue',
   description: 'Socialite',
   age: 35,
@@ -42,7 +43,7 @@ const MrsPeacock = {
 const ColonelMustard = {
   firstName : 'Jack',
   lastName: 'Mustard',
-  color: 'Mustard',
+  color: 'Yellow',
   description: 'Wise old man',
   age: 75,
   image: 'assets/mustard.png',
@@ -178,10 +179,10 @@ const pickKiller = () => {
   mystery.killer = randomSelector(suspects)
   
   //This will change the background color of the card to the one connected to the chosen killer and show the full name of the killer. Feel free to add more things to show about the killer.
-  document.getElementById('killerCard').style.background = mystery.killer.color
+  document.getElementById('killerCard').style.background = mystery.killer.color;
+  document.getElementById('killerImage').src = mystery.killer.image;
   document.getElementById('killerName').innerHTML = `${mystery.killer.firstName} ${mystery.killer.lastName}`
 }
-
 
 const pickWeapon = () => {
   // This will randomly select a weapon. And add that to the mystery object.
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -1,13 +1,11 @@
...
+I tried to fucus on learning more than the design, will make tome to come back for that. Boot-camp life is busy! 
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
 
 ## The problem
-
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+I get stuck in the details. I need to be stricter with my time allocation and ask for help when i need it. When I do, I need to practice formulating my questions in the best way.
```
puj, line 7: Tip:
- Get it working
- Get it working correctly
- Optimize/Refactor/Refine

code/clue.js
```diff
@@ -1,59 +1,227 @@
 // STEP 1 - CREATE OBJECTS FOR ALL THE SUSPECTS, SOMETHING LIKE THIS:
+//
 
-const mrGreen = {
-  firstName: 'Jacob',
+const MrGreen = {
```
puj: Careful with variable naming.  Conventially, variables begin with lowercase letters in Javascript and *classes* (like `Date`) begin with uppercase letters 👍

code/clue.js
```diff
@@ -1,59 +1,227 @@
...
+let weapons = [
+rope,
+knife,
+candlestick,
+dumbbell,
+poison,
+axe,
+bat,
+trophy,
+pistol
```
puj, line 146: What happened to the formatting here?  Why is it so different from the array above?

code/clue.js
```diff
@@ -1,59 +1,227 @@
...
-  mrsWhite
-  // ...  and the rest
+let suspects = [
+  MrGreen,
+  ProfessorPlum,
+  MissScarlet,
+  MrsPeacock,
+  ColonelMustard,
+  MrsWhite
+]
```
puj, line 135: I'm a bit curious why you chose `let` instead of `const` here?  
Unless you expect `suspects` to change to a different value while the webpage is running this can be a constant.

code/clue.js
```diff
@@ -1,59 +1,227 @@
...
-  ).innerHTML = `${mystery.killer.firstName} ${mystery.killer.lastName}`
+  document.getElementById('roomCard').style.background = mystery.room.color
+  document.getElementById('roomName').innerHTML = `${mystery.room}`
 }
 
-// CREATE FUNCTIONS pickWeapon and pickRoom in a similar way.
 
 // STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
-// 'The murder was committed by Jacob Green, in the living room with a rope.'
+const theMystery = document.getElementById('mystery');
```
puj: A variable name describing what kind of element this is could be useful? For example:
`mysteryDiv`
`mysteryButton`
`mysteryHeader`

code/clue.js
```diff
@@ -1,59 +1,227 @@
...
+const revealButton = document.getElementById (`revealButton`);
+const mysteryAnswer = document.getElementById(`mystery`)
```
puj: This is awfully similar to line 206, is there a reason for that?

code/clue.js
```diff
@@ -1,59 +1,227 @@
...
+
+const revealMystery = () => {
+  if (mystery.killer.firstName === undefined
+    || mystery.weapons.name === undefined
+    || mystery.room === undefined)
```
puj: This logic allows you to click the killer and weapon card and press reveal without clicking the room card.  

This is because of line 169.

Consider the following idea
```Javascript
const mystery = {
  killer: undefined,
  weapons: undefined,
  room: undefined
}
```

Try to understand why this would make your page behave differently 👍

code/clue.js
```diff
@@ -1,59 +1,227 @@
...
+const theMystery = document.getElementById('mystery');
+const revealButton = document.getElementById (`revealButton`);
+const mysteryAnswer = document.getElementById(`mystery`)
+
+const revealMystery = () => {
+  if (mystery.killer.firstName === undefined
+    || mystery.weapons.name === undefined
+    || mystery.room === undefined) 
+    {
+    mysteryAnswer.innerHTML = `No peeking! Guess and click each card first!`
```
puj, line 215: I like this message 😆

code/clue.js
```diff
@@ -1,59 +1,227 @@
...
+  else {
+theMystery.innerHTML =` The murder was committed by ${mystery.killer.firstName} ${mystery.killer.lastName}, 
+in the ${mystery.room} with a ${mystery.weapons.name}`
+  }
+};
+
+document.getElementById(`revealButton`).onclick = revealMystery;
+document.getElementById('killerCard').onclick = pickKiller;
+document.getElementById('weaponCard').onclick = pickWeapon;
+document.getElementById('roomCard').onclick = pickRoom;
```
puj, line 226: I love that ALL of these are grouped together.  I feel like I know where all of your event handling happens.  

Also, if I did this for myself, I would know exactly where to add a new `onclick` handler, should I create a new one in the future!

code/index.html
```diff
@@ -5,16 +5,20 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <link rel="stylesheet" href="style.css" />
+    <link href="https://fonts.googleapis.com/css?family=Abril+Fatface|Coda+Caption:800|Lacquer&display=swap" rel="stylesheet">
```
puj, line 8: Spooooky! 👻

README.md
```diff
@@ -1,13 +1,11 @@
 # Cluedo - Mystery Randomizer
+This is my second week into learning js, this project was supposed to teach me more about loors and arrays. 
+I tried to fucus on learning more than the design, will make tome to come back for that. Boot-camp life is busy!
```
puj, line 3: My personal opinion is focus on the main topic for the week.  Functionality is beauty this week!

puj (approved):
Good job!  I'm so glad that you focused on solidifying the core concepts we focused on this week.  We'll be building on these and tying together with other more complex ideas as well as revisiting design.  

Plenty of everything to come!  I'm excited to see where you take your newfound knowledge. 

Feel free to comment in response to the comments if you want more feedback.

code/clue.js
L-Lisa: If you ment the caps, my head thinks it needs to capitalise Names and not html, I need to forget about that!

code/clue.js
L-Lisa: I tried to follow your example but could not get it to work, the text would not show. 
Instead, I added " " in the if statement too, and that worked, I didn't really understand how that worked and not the undefined option but will figure it out now.

code/clue.js
L-Lisa: Thank you, trying to embrace order :)

L-Lisa:
Thank you for your feedback, if you have time I have two questions. 
When going through your responses I corrected the code and repushed it, how does that work, should I make a new pull request then?
Secondly, you suggested I write undefined as the value of the mystery object items, I couldn't get it to work but instead added the same "" to the if statement and that seems to work. Was that an ok solution? 
Have a great day!

L-Lisa:
ohhh and it wants me to leave a final comment here, I'll just copy the one I left above thinking that was the right place. Im really pleased were getting some practice on this system!!! :)

"Thank you for your feedback, if you have time I have two questions.
When going through your responses I corrected the code and repushed it, how does that work, should I make a new pull request then?
Secondly, you suggested I write undefined as the value of the mystery object items, I couldn't get it to work but instead added the same "" to the if statement and that seems to work. Was that an ok solution?
Have a great day!

PR #82, opened 2020-02-15
Project 5 - Ida

idaabra:
I have completed the assignment and also a stretch goal of putting images to all of the objects. Some images doesn't work but I don't feel to put more effort into that a the moment. I struggle when I want to do something that I know I've done in previous projects but don't really remember how to. I feel like I never really have time to look that up. I guess it can be a matter of me wanting to do more than I am capable of at the moment. Still pleased that I figured this assignment out fairly quickly and also that I think I understand it also.


code/index.html
```diff
@@ -1,37 +1,42 @@
...
 
-    <button>REVEAL THE CRIME</button>
+<body>
+  <section class="cards">
+    <article class="killer-deck" id="killerCard">
+      <p>KILLER</p>
+      <p id="killerName"></p>
+      <p id="killerAge"></p>
+      <p id="killerOccupation"></p>
+      <img id="killerImage" src="" height="100" />
```
ericaglimsholt, line 19: Try to add the styling parts in the css instead of in the html :)

ericaglimsholt (approved):
Hi Ida! Thanks for submitting your project. 
I will change the status from late to done and approved :)

PR #86, opened 2020-04-16
Jihyun An Cluedo 

jihyeona:


code/clue.js
```diff
@@ -9,51 +7,179 @@ const mrGreen = {
...
+  mystery.room = randomSelector(rooms)
+  document.getElementById('roomCard').style.background = mystery.killer.color;
+  document.getElementById('roomName').innerHTML = mystery.room;
+}
+document.getElementById('roomCard').addEventListener('click', pickRoom)
+
+const revealMystery = () => {
+  document.getElementById('revealAnswer').innerHTML = `The murder was committed by ${mystery.killer.firstName} ${mystery.killer.lastName}, in the ${mystery.room} with a ${mystery.weapon.name}.`
+}
+document.getElementById('revealAnswer').addEventListener('click', revealMystery)
```
puj: You shouldn't be able to reveal the crime until all of the cards are chosen.

puj (changes requested):
Let me know when you fix the reveal mystery function and we can get this approved

code/clue.js
jihyeona: I have made the change!

HIPPIEKICK (approved):

PR #93, opened 2020-09-17
Project week 5 Cluedo - Karolin Albinsson

karolinalbinsson:
For this project I have tried to think about the looks of my code as I have gotten some negative feedback on that point in previous projects. I hope I have improved on that point. 
I aimed for black level in this project.


code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+    document.getElementById(loadID.id).style.opacity = 1;
+    setTimeout(function () {
+      document.getElementById(loadID.id).style.opacity = 0;
+      const increasedRoom = increaseChanceForRoom(mystery.weapon);
+      mystery.room = randomSelector(increasedRoom);
+      document.getElementById('roomName').innerHTML = mystery.room;
+      roomPicked = true;
+      checkScroll();
+    }, 2000);
+  } else(showInfoMessage("pickRoom"));
```
emeliesv, line 259: This is a great function, but when I visited your page the 'else' is never triggered.

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+    document.getElementById("weaponInfoMessage").style.opacity = 1;
+    setTimeout(function () {
+      document.getElementById("weaponInfoMessage").style.opacity = 0;
+    }, 2000);
+  } else if (callerFunctionName === "revealMystery") {
+    document.getElementById("mysteryInfoMessage").style.opacity = 1;
+    setTimeout(function () {
+      document.getElementById("mysteryInfoMessage").style.opacity = 0;
+    }, 2000);
+  }
```
emeliesv, line 278: As I wrote above, this function is awesome! It's a great idea to have something that triggers when the user haven't picked a card. However, I am missing an info-message for the room-card. Since that likely is the last card to be chosen, maybe this function could have been triggered by a button? That way the user would be guided through picking the cards.

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+    setTimeout(function () {
+      document.getElementById(loadID.id).style.opacity = 0;
+      mystery.weapon = randomSelector(weaponsIncreased);
+      document.getElementById('weaponName').style.background = mystery.weapon.color;
+      document.getElementById('weaponName').innerHTML = `Type: ${mystery.weapon.name}`;
+      document.getElementById('weaponWeight').innerHTML = `Weight: ${mystery.weapon.weight}`;
+      weaponPicked = true;
+      checkScroll();
+    }, 2000);
+  } else if (showInfoMessage("pickWeapon"));
```
emeliesv, line 245: Maybe it's enough with only an else here? :)

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+  toggleGameBoard();
+  shuffleFavoriteWeapon();
+}
+
+const revealMystery = () => {
+  if (killerPicked && weaponPicked && roomPicked) {
+    const killer = mystery.killer.firstName + " " + mystery.killer.lastName;
+    const weapon = mystery.weapon.article + " " + mystery.weapon.name;
+    const room = mystery.room;
+    const mysteryString = (`
```
emeliesv, line 322: I think maybe this formatting wasn't intentional, but since it's here perhaps you could reformat this so it's easier to read? So instead of having this message split from line 322-330 it would read 
`const mysteryString = (It was ${killer} using ${weapon} in the ${room}!);` (But with the backticks you're already using of course) on line 322 instead.

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+                using ${
+                  weapon
+                } in the ${
+                  room
+                }!`);
+    document.getElementById('mystery-reveal').innerHTML = mysteryString;
+  } else(showInfoMessage("revealMystery"));
+}
+
+const restartGame = () => {
```
emeliesv, line 335: This was really neat! 💯

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+      const weaponID = mystery.killer.favoriteWeapon;
+      const weaponName = weapons.find(weapon => weapon.id === weaponID).name;
+      document.getElementById('killerFavoriteWeapon').innerHTML = (`Preferred weapon: ${weaponName}`);
+      document.getElementById('killerAge').innerHTML = (`Age: ${mystery.killer.age}`);
+      document.getElementById('killerOccupation').innerHTML = (`Occupation: ${mystery.killer.occupation}`);
+      document.getElementById('killerDescription').innerHTML = (`Description: ${mystery.killer.description}`);
+      killerPicked = true;
+      checkScroll();
+    }, 2000);
+  }
```
emeliesv, line 229: I don't know if this was intentional or not, but for some reason I couldn't reshuffle your killerdeck once I've clicked on the killer-deck. Maybe it's because you removed the randomSelector from suspects? However, it doesn't affect the game itself :)

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+  const y = window.matchMedia("(max-width: 769px)");
+  if (x.matches) {
+    setTimeout(function () {
+      window.scrollBy(0, window.innerHeight / 1.3);
+    }, 1500);
+  } else if (y.matches) {
+    setTimeout(function () {
+      window.scrollBy(0, window.innerHeight / 1.5);
+    }, 1500);
+  }
```
emeliesv, line 360: Nice! 👍

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+  const preferredRoom = weaponObject.likelyFoundIn;
+  const increaseChance = Math.floor(Math.random() * 10) + 1;
+  for (i = 0; i < increaseChance; i++) {
+    roomsIncreased.push(preferredRoom);
+  }
+  roomsIncreased = roomsIncreased.concat(rooms);
+  return roomsIncreased;
+}
+
+const startGame = () => {
```
emeliesv, line 311: This was a nice touch that gave the page a cleaner look! Sweet 👍

code/style.css
```diff
@@ -1,6 +1,35 @@
...
+}
+
+@font-face {
+  font-family: Futura-Extra-Bold;
+  src: url(./fonts/Futura\ Extra\ Bold.otf);
+}
+
+.header {
+  min-height: 20vh;
+  background: black;
```
emeliesv, line 22: Perhaps stick to hex codes?

emeliesv (commented):
Great work! I learned a lot from reading your code, and I really like the checkScroll() you made! I left only one comment for the CSS, which concerns wether you should use hex-codes when picking colors. I don't know if you have to use hex-codes or not, but at the moment you're using hexcode for the body and the loader-border but colornames for the rest (might be nothing to think about though!)

At the moment the suspect is only random when restaring the game, so the shuffle-function for the suspects are gone (however, perhaps that is a question of interpretation in the blue requirements since it doesn't say anything about the need for shuffling the deck once you've picked a killer?). There is one problem with indentation on line 322 I noticed that you should take a look at.

I think maybe it would be nice with a few comments explaining some of the code to easier understand what the functions do, but as I said: Great job! I was very impressed! 🥇

emeliesv (changes requested):
I've posted this comment again since I missed choosing Approve/Request changes before I submitted my review!

Great work! I learned a lot from reading your code, and I really like the checkScroll() you made! I left only one comment for the CSS, which concerns wether you should use hex-codes when picking colors. I don't know if you have to use hex-codes or not, but at the moment you're using hexcode for the body and the loader-border but colornames for the rest (might be nothing to think about though!)

At the moment the suspect is only random when restaring the game, so the shuffle-function for the suspects are gone (however, perhaps that is a question of interpretation in the blue requirements since it doesn't say anything about the need for shuffling the deck once you've picked a killer?). There is one problem with indentation on line 322 I noticed that you should take a look at.

I think maybe it would be nice with a few comments explaining some of the code to easier understand what the functions do, but as I said: Great job! I was very impressed! 🥇

code/clue.js
puj: It seems like it was intentional
`if (gameStarted && !killerPicked) {`

Good testing @emeliesv

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
-// With a killer, a weapon and a room.
-// The values will be set later.
+const mystery = {
+  killer: null,
+  weapon: null,
+  room: null
+};
+
+const pickKiller = (loadID) => {
+  if (gameStarted && !killerPicked) {
```
puj, line 212: One idea here could be to 
```Javascript
if( !gameStarted || killerPicked) {
   return;
}

// Now you know the game is started and the killer is not picked
```

Inverting your logic would allow you to not have the entire function in the `if` 👍

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+    document.getElementById(loadID.id).style.opacity = 1;
+    setTimeout(function () {
+      document.getElementById(loadID.id).style.opacity = 0;
+      mystery.killer = randomSelector(suspects);
+      document.getElementById('killerName').style.background = mystery.killer.color;
+      document.getElementById('killerImage').src = mystery.killer.image;
+      document.getElementById('killerName').innerHTML = (`${mystery.killer.firstName} ${mystery.killer.lastName}`);
+      const weaponID = mystery.killer.favoriteWeapon;
+      const weaponName = weapons.find(weapon => weapon.id === weaponID).name;
```
puj, line 221: Cooooool!

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+      document.getElementById('killerFavoriteWeapon').innerHTML = (`Preferred weapon: ${weaponName}`);
+      document.getElementById('killerAge').innerHTML = (`Age: ${mystery.killer.age}`);
+      document.getElementById('killerOccupation').innerHTML = (`Occupation: ${mystery.killer.occupation}`);
+      document.getElementById('killerDescription').innerHTML = (`Description: ${mystery.killer.description}`);
+      killerPicked = true;
```
puj, line 226: I'm interested to know why you chose to create a variable for this rather than relying on the `mystery` data.  

Another alternative is to create a function like
```javascript
const isKillerPicked = () => mystery.killer !== null;
```
Or something similar

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+  }
+}
 
-// This function will be invoked when you click on the killer card.
-const pickKiller = () => {
-  // This will randomly select a killer from the suspects. And add that to the mystery object.
-  mystery.killer = randomSelector(suspects)
+const pickWeapon = (loadID) => {
+  if (killerPicked) {
+    const weaponsIncreased = increaseChanceForFavWeapon(mystery.killer);
```
puj, line 234: Love this name `increaseChanceForFavWeapon`

puj: `weaponsIncreased ` also very good naming

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+    }, 2000);
+  }
+}
 
-// This function will be invoked when you click on the killer card.
-const pickKiller = () => {
-  // This will randomly select a killer from the suspects. And add that to the mystery object.
-  mystery.killer = randomSelector(suspects)
+const pickWeapon = (loadID) => {
+  if (killerPicked) {
```
puj, line 233: Once again, the logic of
```javascript 
if(!killerPicked){
  showInfoMessage("pickWeapon")
}else{
  ...
}
``` 
Could be more readable

code/clue.js
puj: This confused me a bit because I initially read it as 
```javascript
} else (showInfoMessage("pickRoom")) {
 ...
}
```
As if `showInfoMessage` should return a value, but it doesn't.
As @emeliesv mentioned, there is really no else.  Javascript will interpret this as 
```javascript 
} else (showInfoMessage("pickRoom")){
 ;
}
```
Which calls the function `showInfoMessage` when checking if the `else` is `true/false` but disregards the result.  The end result is the same since you are showing the message though.

Restructure it a bit to be more clear 👍

code/clue.js
puj: Some comments could help a lot at first glance here.

Also, consider simplifying this function by creating a `displayHint` function.  Maybe something like

```javascript
const displayHint = (hintId) =>{
  document.getElementById(hintId).style.opacity = 1;
  setTimeout(function () {
      document.getElementById(hintId).style.opacity = 0;
    }, 2000);
}
```

It might make this function more readable

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+      document.getElementById("mysteryInfoMessage").style.opacity = 0;
+    }, 2000);
+  }
+}
+
+const shuffleFavoriteWeapon = () => {
+  suspects.forEach(suspect => {
+    const selectedRandomWeapon = randomSelector(weapons).id;
+    suspect.favoriteWeapon = selectedRandomWeapon;
+  });
```
puj, line 285: Nice

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+  let weaponsIncreased = new Array();
+  const preferredWeapon = killerObject.favoriteWeapon;
+  const weaponObject = weapons.find(weapon => weapon.id === preferredWeapon);
+  const increaseChance = weaponObject.increasedChanceOfDraw;
+  for (i = 0; i < increaseChance; i++) {
+    weaponsIncreased.push(weaponObject);
+  }
+  weaponsIncreased = weaponsIncreased.concat(weapons);
+  return weaponsIncreased;
+}
```
puj, line 298: Really beautiful.  Only one small suggestion `const increasedChance` instead of `const increaseChance`.  Most likely you intended that already.

code/clue.js
```diff
@@ -1,59 +1,361 @@
...
+
+const increaseChanceForRoom = (weaponObject) => {
+  let roomsIncreased = new Array();
+  const preferredRoom = weaponObject.likelyFoundIn;
```
puj, line 302: Continuously impressed by your naming `likelyFoundIn` 💯

code/clue.js
puj: 100% agree

code/clue.js
puj: This looks like an auto-formatter setting, I agree with you @emeliesv

code/clue.js
puj: Agreed.  A comment to explain to the reader what this function's intention is would be great.

code/style.css
puj: Consistency

puj:
This was a very impressive project.  Good understanding of core concepts and a lot of successful exploration beyond.  Continue with your naming conventions, professional quality and very well thought out. Don't underestimate that skill! 💯 

@emeliesv Very insightful comments and good attention to detail, nicely done!

puj (approved):

PR #100, opened 2020-09-18
Project Cluedo - Ellen

ellenbjo:
I was aiming for the red level with this weeks project, but unfortunately I couldn't figure out how to solve the shuffelfavouriteweapon function in the way I wanted. But overall I'm happy about the project- I've learned a lot this week and had a lot of fun doing the project! I'm proud of getting the functions to work together and that I finally solved the loading animation to show the way I wanted it to,


code/index.html
```diff
@@ -8,29 +8,61 @@
...
   </head>
   <body>
+  
+    <header>
+      <div class="heading-container">
+        <h1>CLUEDO</h1>
+        <img class="icon" src="./assets/detective.svg" alt="image of a magnifying glass">
+      </div>
+      <div class="intro-text-container">
+      <P>Click on the cards to select a killer, a weapon and a room.</P>
```
VanessaSue27: Just a small detail here, this `<p>` element should have indentation, two spaces into the `<div>` element.
And not sure this makes a difference, but I usually see `<p>` not `<P>`

code/index.html
```diff
@@ -8,29 +8,61 @@
...
     </section>
 
-    <h1 id="mystery"></h1>
+    <section class="reveal-section">
+      <div class="reveal-mystery-text">
+        <p id="mysteryAnswer"></p>
+      </div>
+
+      <button class="reveal-button" id="revealButton" onclick="revealMystery()">REVEAL THE CRIME</button>
+      <button class="new-game-button" id="newGameButton" onclick="location.reload()">NEW GAME</button>
```
VanessaSue27, line 57: Thanks to you I now learned about `location.reload()` 👌

code/index.html
```diff
@@ -8,29 +8,61 @@
...
+    </section>
 
-    <button>REVEAL THE CRIME</button>
+    <footer>
+      <div class="footer-text"> 
+        Icons made by <a href="https://www.flaticon.com/authors/freepik" title="Freepik">Freepik</a> and <a href="https://www.flaticon.com/authors/dinosoftlabs" title="DinosoftLabs">DinosoftLabs</a> from <a href="https://www.flaticon.com/" title="Flaticon">www.flaticon.com</a> 
+      </div>
+    </footer>
 
     <script src="clue.js"></script>
```
VanessaSue27: On line 7 (where the CSS file is linked) and also on this line, I understand it's better that we start the link with ./
Like this: `src="./clue.js"`
So that it knows it's the clue.js file in the same folder.

code/style.css
```diff
@@ -1,58 +1,290 @@
...
+  width: 25%;
+  height: 470px;
   background: midnightblue;
-
   text-align: center;
   padding-top: 30px;
   border-radius: 16px;
+  box-shadow: 0 0 5px 5px #822c5b;
+  cursor: pointer;
 }
```
VanessaSue27: Card decks all have the same styling, except only for the background-color.
Maybe there is a way to optimize this? 👩‍💻

code/style.css
```diff
@@ -1,58 +1,290 @@
...
+  margin: auto;
+  padding: 20px;
+}
+
+.killer-deck .card-image {
+  border-radius: 50%;
+}
+
+footer {
+  display: flex;
```
VanessaSue27, line 137: The footer seems to be just one line of text with some links included.
Is it necessary for it to be a flexbox?
Maybe can it be a simple `<p>` element with `<a>` in between for the links, and center with `text-align` ?

code/style.css
```diff
@@ -1,58 +1,290 @@
...
+  }
+
+  .cards {
+    display: flex;
+    flex-wrap: wrap;
+    padding-bottom: 900px;
+    background: transparent;
+    width: 80%;
+    height: 700px;
+    background-color: #f3f3f3;
```
VanessaSue27: There's a `background: transparent` and a `background-color` both declared here for `.cards` in tablet version.
Maybe the intention is to make the background transparent when it switches to tablet view, but it doesn't happen because there's a background-color also set here?

code/style.css
```diff
@@ -1,58 +1,290 @@
...
+
+  .killer-deck {
+    width: 100%;
+    margin-bottom: 10px;
+  }
+
+  .weapon-deck {
+    width: 100%;
+    margin-bottom: 10px;
+  }
```
VanessaSue27: Maybe these two decks can be put under the same style declaration? Since styling is the same.

code/clue.js
```diff
@@ -1,59 +1,408 @@
-// STEP 1 - CREATE OBJECTS FOR ALL THE SUSPECTS, SOMETHING LIKE THIS:
+const killerCard = document.getElementById("killerCard");
+const weaponCard = document.getElementById("weaponCard")
+const roomCard = document.getElementById("roomCard")
```
VanessaSue27: Consitency with semi-colons, they appear in line 1, but not on 2 and 3

code/clue.js
```diff
@@ -1,59 +1,408 @@
...
+const mystery = {
+  killer: undefined,
+  weapon: undefined,
+  room: undefined 
+};
 
-// CREATE AN OBJECT THAT KEEPS THE MYSTERY.
-// With a killer, a weapon and a room.
-// The values will be set later.
+// FUNCTIONS
```
VanessaSue27, line 304: I understood the flow of your code is as follows:
First: functions are called with onclick (lines 390-392)
Second: the startAnimation functions start to run
Third: then we have the pickKiller, pickWeapon and pickRoom functions

Maybe it would be nicer to re-organize the order of the functions so we can more easily follow the flow of what is happening first, second and third?
Right now the onclick functions are at the bottom, maybe those should come first? 
Maybe put the animation functions before the pickKiller, pickWeapon and pickRoom functions also? Just to have a nice order of which functions are coming into play first, followed by the next ones accordingly.

code/clue.js
```diff
@@ -1,59 +1,408 @@
...
-// This function will be invoked when you click on the killer card.
 const pickKiller = () => {
-  // This will randomly select a killer from the suspects. And add that to the mystery object.
-  mystery.killer = randomSelector(suspects)
+  mystery.killer = randomSelector(suspects);
+
+  const killerImage = document.getElementById("killerImage");
+  const favouriteWeapon = document.getElementById("favouriteWeapon");
+  const killerAge = document.getElementById("killerAge");
+  const killerName = document.getElementById("killerName");
```
VanessaSue27: These elements which are the connection to HTML (all the ones using `document.getElementByID`) are declared twice: once here in the pickKiller function, and also once more in the animation function (lines 350-353).
Same happens in pickWeapon and pickRoom combined with startAnimation2 and startAnimation3
Maybe there is a way they don't need to be declared twice?

code/clue.js
```diff
@@ -1,59 +1,408 @@
...
+const startAnimation1 = () => {
+  const killerImage = document.getElementById("killerImage");
+  const favouriteWeapon = document.getElementById("favouriteWeapon");
+  const killerAge = document.getElementById("killerAge");
+  const killerName = document.getElementById("killerName");
+
+  killerImage.style.display = "none";
+  favouriteWeapon.style.display = "none";
+  killerAge.style.display = "none";
+  killerName.style.display = "none";
```
VanessaSue27, line 352: This section contains all the killer data, maybe can a `<section>` be created in HTML to contain all this data and then change that section to `display: none;` instead of doing it for every line separately?
Same for lines 369-371 and 381-382

Also, this initial `display: none;` styling, can it be set like that in the CSS since the start (instead of here in the JS file)
and then changed to `display: block; `when the pickKiller, pickWeapon and pickRoom functions are called?

code/clue.js
```diff
@@ -1,59 +1,408 @@
...
-  document.getElementById(
-    'killerName'
-  ).innerHTML = `${mystery.killer.firstName} ${mystery.killer.lastName}`
-}
+// FINAL FUNCTION
 
-// CREATE FUNCTIONS pickWeapon and pickRoom in a similar way.
+const revealMystery = () => {
+  const answerMystery = document.getElementById("mysteryAnswer")
+  if (mystery.killer === undefined || mystery.weapon === undefined || mystery.room === undefined) {
```
VanessaSue27, line 386: Nice adding this condition to make sure all cards are clicked before revealing the mystery! 😎

code/clue.js
```diff
@@ -1,59 +1,408 @@
...
+  const killerAge = document.getElementById("killerAge");
+  const killerName = document.getElementById("killerName");
+
+  killerImage.style.display = "none";
+  favouriteWeapon.style.display = "none";
+  killerAge.style.display = "none";
+  killerName.style.display = "none";
+  document.getElementById("loadingAnimation1").style.display = "block";
+
+  setTimeout(pickKiller, 1500)
```
VanessaSue27: Super nice that you kept trying until it worked to implement the loaders! 🎉

code/clue.js
```diff
@@ -1,59 +1,408 @@
...
 
-// CREATE FUNCTIONS pickWeapon and pickRoom in a similar way.
+const revealMystery = () => {
+  const answerMystery = document.getElementById("mysteryAnswer")
+  if (mystery.killer === undefined || mystery.weapon === undefined || mystery.room === undefined) {
+      answerMystery.innerHTML = ("Please choose a killer, a weapon and a room.");
+    } else {
+      answerMystery.innerHTML = (`The murder was committed by ${mystery.killer.firstName} ${mystery.killer.lastName}, in the ${mystery.room.name} with a ${mystery.weapon.name}.`);
+      document.getElementById("revealButton").style.display = "none";
+      document.getElementById("newGameButton").style.display = "block";
```
VanessaSue27, line 391: Also really nice that you include a new game button so user can try again 👌

VanessaSue27 (changes requested):
Great job Ellen! Your code is very easy to follow and understand.
I love that you took the time to also include images for weapons and rooms (nice icons!) and that they also change background color.
Site is responsive! YAY!!!

Some small things I noticed:
- The .gitignore file is missing (to avoid having those .DS_Store files)
- After being clicked for the first time, cards can still be clicked more times. 
This was not really in the requirements, but maybe a nicer experience when the card is revealed once and then it can't be clicked again.

Changes Requested:
Just the small indentation I pointed out, and missing semi-colons.
I've been told in previous projects that this needs to be changed before the PR can be approved, but for the rest, nothing seems to be crucial.

It was my pleasure reviewing your code! /Vanessa 🤗

README.md
```diff
@@ -1,13 +1,19 @@
...
+The focus for the project is to practice objects and arrays and also to manipulate the DOM with JavaScript.
 
-Replace this readme with your own information about your project.
+## The problem
 
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+This week was all about learning how to work with arrays, objects and how to get functions working together. 
 
-## The problem
+I've spent a lot of hours trying to figure out how to shuffle the favourite weapon, but unfortunately I couldn't make it work the way I wanted it to. BUT this really motivated me to learn more about foreach loops and arrays and resulted in many hours of googling and researching.
```
puj: What a fantastic reflection! 👏

code/clue.js
puj: Really helpful comment @VanessaSue27 
Highlighting what was difficult for you as the reader and providing a potential solution. 

Usually the discussion would continue with a proposed way forward and advantages/disadvantages to changing/preserving the approach. 👍

code/clue.js
puj: "Maybe there is a way they don't need to be declared twice?"
Perfect

code/clue.js
```diff
@@ -1,59 +1,408 @@
...
+  const favouriteWeapon = document.getElementById("favouriteWeapon");
+  const killerAge = document.getElementById("killerAge");
+  const killerName = document.getElementById("killerName");
+
+  killerImage.style.display = "block";
+  favouriteWeapon.style.display = "block";
+  killerAge.style.display = "block";
+  killerName.style.display = "block";
+  killerCard.style.background = mystery.killer.color;
+  document.getElementById("loadingAnimation1").style.display = "none";
```
puj: `loadingAnimation1` could be something more specific

 `document.getElementById("loadingAnimation1").style.display = "none";` 
This line of code by itself would be very difficult to understand in relation to what will change on the page

code/clue.js
puj: Interesting, so making a container that hides, rather than hiding each element?

code/clue.js
puj: Agreed! Well done!

code/index.html
puj: @VanessaSue27 Good chance for a code suggestion 
```html
<div class="intro-text-container">
   <p>Click on the cards to select a killer, a weapon and a room.</p>
</div>
```

code/index.html
puj: Thanks for sharing this learning!

code/style.css
puj: Maybe it overwrites?

puj:
@ellenbjo Vanessa provided you with a really good code review and I don't have much more to add

She mentioned:

> - After being clicked for the first time, cards can still be clicked more times.
> This was not really in the requirements, but maybe a nicer experience when the card is revealed once and then it can't be clicked again.

A possible alternative, is that if you allow the cards to be clicked multiple times after the mystery is revealed, that you hide the reveal mystery text once a card has been clicked.

Say I click Reveal Mystery and I see "The murder was committed by Blanche White, in the Ballroom with a rope."
If I then click the weapon card, maybe that text should dissappear? 

I also agree with the Changes Requested

ellenbjo:
thank you @VanessaSue27 and @puj for your feedback! I really appreciate it! :)  I will do the changes requested and go through my code again.

puj (approved):

PR #103, opened 2020-09-18
Week 5 cluedo

monkeybusiness101:


code/clue.js
```diff
@@ -3,33 +3,181 @@
...
+  rope,
+  knife,
+  candlestick,
+  dumbbell,
+  poison,
+  axe,
+  bat,
+  trophy,
+  pistol
+]
```
jespa374, line 163: Not sure if it's preferred to write array items vertically or horizontally but I think it saves a lot of space if you write as a row if there are not so many items like in the rooms array.

code/clue.js
```diff
@@ -41,19 +189,73 @@ const randomSelector = array => {
 // With a killer, a weapon and a room.
 // The values will be set later.
 
+let mystery = {
+  killer:'',
+  weapon:'',
+  room:''
+}
```
jespa374: Any reason for using let instead of const? It's recommended to use const as much as possible if there is not a specific reason for using let.

code/clue.js
```diff
@@ -41,19 +189,73 @@ const randomSelector = array => {
...
 // This function will be invoked when you click on the killer card.
+
 const pickKiller = () => {
   // This will randomly select a killer from the suspects. And add that to the mystery object.
   mystery.killer = randomSelector(suspects)
+  
+  // Shuffle weapon function
+  shuffleFavouriteWeapon = () => {
+    mystery.killer.favouriteWeapon = randomSelector(weapons); 
+  };
```
jespa374: You are using semicolons here but everywhere else you are not using semicolon. Be consistent.

code/clue.js
```diff
@@ -41,19 +189,73 @@ const randomSelector = array => {
...
+    shuffleFavouriteWeapon();
+    console.log(mystery.killer.favouriteWeapon.name)
```
jespa374, line 210: No need of this console.log right? Maybe you can remove it.

code/clue.js
```diff
@@ -41,19 +189,73 @@ const randomSelector = array => {
...
 
   // This will change the background color of the card to the one connected to the chosen killer and show the full name of the killer. Feel free to add more things to show about the killer.
   document.getElementById('killerCard').style.background = mystery.killer.color
+  document.getElementById('killerImage').src = mystery.killer.image
   document.getElementById(
     'killerName'
   ).innerHTML = `${mystery.killer.firstName} ${mystery.killer.lastName}`
+  document.getElementById('killerFavoriteWeapon').innerHTML = `${mystery.killer.favouriteWeapon.name}`
 }
+document.getElementById('killerCard').addEventListener('click', pickKiller)
```
jespa374, line 220: Another way to write this would be `document.getElementById('killerCard').onclick = pickKiller`

code/index.html
```diff
@@ -7,13 +7,20 @@
...
+  <! -- BODY -->
   <body>
     <section class="cards">
       <article class="killer-deck" id="killerCard">
         <p>KILLER</p>
         <p id="killerName"></p>
         <p id="killerAge"></p>
         <p id="killerOccupation"></p>
+        <p id="killerFavoriteWeapon"></p>
         <img id="killerImage" src="" height="100" />
```
jespa374: Add an alt attribute to the img in case the picture isn't visible and for accessibility

jespa374 (approved):
Great job Max! :) You have done the blue level steps and also achieved red level with the favourite shuffle weapon. Congrats! I think you write clean code and the comments really help the viewer. Apart from the points mentioned in the code you can probably style the page more and also try to make the page more responsive in case you want to improve it but overall it's well done  👍

code/clue.js
EmOlsz: I would agree with Max over here - it's far more clean to format array in a away that single element has separate line

code/clue.js
EmOlsz: Good point Jessica!
Moreover, super small thing, but I would suggest you to initialise your mystery object with null values rather than empty strings. It's about sending proper information for other developers : null value indicates that probably we will replace these values with something else, but without saying what type will that new value have for now. Initialising your object with empty string values ' ' could be misunderstood by some one in a way that your properties will later on by replaced by new strings only. As a sum up, this would be better :
```
const mystery = {
  killer: null,
  weapon: null,
  room: null,
}
```

code/clue.js
EmOlsz: Once again, good job Jessica!
One more thing to point out - in line 208 there is a declaration of variable without any keyword `let` or `const`. Because of that you've created global variable `shuffleFavouriteWeapon`. Briefly speaking, of you would create `const shuffleFavouriteWeapon = ...`, it would be scoped to in `pickKiller` function only and unavailable from outer scope. Because you omitted `const`, your variable somehow "slipped out" from `pickKiller` function.
This is huge antipattern, we should never (well, in 99.99% cases being honest) create global variables. Please, fix that asap

code/clue.js
EmOlsz: 👍

code/clue.js
```diff
@@ -3,33 +3,181 @@
 const mrGreen = {
   firstName: 'Jacob',
   lastName: 'Green',
-  color: 'green',
+  color: 'Green',
```
EmOlsz, line 6: `color` property of suspects objects should be written lowercase - CSS expect from us lowercase values of colours

code/clue.js
EmOlsz: Yes, but it's alternative, you can choose whatever way you want

code/clue.js
```diff
@@ -41,19 +189,73 @@ const randomSelector = array => {
...
 
-// CREATE FUNCTIONS pickWeapon and pickRoom in a similar way.
 
 // STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
 // 'The murder was committed by Jacob Green, in the living room with a rope.'
+
+
+const revealMystery = () => {
+  document.getElementById('mystery').innerHTML = `The murder was committed by ${mystery.killer.firstName} ${mystery.killer.lastName}, in the ${mystery.room} with a ${mystery.weapon.name}`
+}
```
EmOlsz, line 259: You forgot to block suspect, weapon and room cards functionality over here - even after revealing crime I was able to change suspect card

code/index.html
EmOlsz: Very nice Jessica

code/style.css
```diff
@@ -1,6 +1,15 @@
 body {
   margin: 0;
   font-family: "Courier New", Courier, monospace;
+  
+
```
EmOlsz, line 5: Unnecessary extra blank line

code/style.css
```diff
@@ -9,20 +18,26 @@ p {
...
 }
 
+/* CARDS BACKGROUND */
 .cards {
   margin: 50px auto;
   width: 90%;
   height: 600px;
-  background: #ccc;
+  background: #093028;  /* fallback for old browsers https://cssgradient.io/gradient-backgrounds/ */
+  background: -webkit-linear-gradient(to right, #237A57, #093028);  /* Chrome 10-25, Safari 5.1-6 https://cssgradient.io/gradient-backgrounds/*/
```
EmOlsz, line 27: Very good Max :)

EmOlsz (changes requested):
Congrats Max, good job this week :) Take a look on both Jessica's and my comments and fix stuff we pointed out. I would especially like you to change unwanted global variable (there is a comment in code about that).
After changes are applied, I will approve your project!

code/clue.js
monkeybusiness101: fixed

code/clue.js
monkeybusiness101: Thank you for that Maks! I added const to the function. I also moved the function out of pickKiller. I don't know if that was what you wanted but perhaps makes the function more useful?

code/index.html
monkeybusiness101: fixed

monkeybusiness101:
@EmOlsz Hi Maks! I have fixed the global variable and moved it out of pickKiller. I have also fixed, indentation, img alt, changed mystery object to null

code/clue.js
EmOlsz: Probably it is, good idea :)

EmOlsz (approved):

PR #110, opened 2020-09-18
Claire Caudwell Cluedo Murder Mystery Picker

ClaireCaudwell:
Hiya! 

This was a project that was both fun and challenging. I enjoyed making the cards clickable, making the reveal mystery button work to show the reveal the mystery sentence, making the loader work and styling my page. I managed to get to the red level, but after that I gave up and want to achieve the black level at a later date.

https://condescending-neumann-78ee0c.netlify.app/


README.md
```diff
@@ -1,13 +1,32 @@
...
+4. In relation to that I also learnt how to use classList to add a CSS class to a certain HTML element in order for my animation to run. Before the animation was run it was a static image of a magnify glass, but once the function was called the classList added my animation class for that img and it made the animation run.
+5. In general when it comes to javaScript I feel like I am getting used to the syntax and how things are worrking together both in the javascript file and the seperate HTML and CSS files.  
+6. Last but not least I learnt from this project that you always have to be thinking about points of execution and which functions relate to that. For example when the user clicks on the card what get's called and why. To be methodical when writing javaScript and to console.log more inside the code to see what is being returned in the console.log in dev tools.
```
AxelHammarback: This is just lovely to read – well done! Reading through these, I jotted down at least three things which it sounds like I need to learn more about 😄

code/clue.js
```diff
@@ -1,59 +1,304 @@
...
 const suspects = [
   mrGreen,
-  mrsWhite
-  // ...  and the rest
-]
+  professorPlum,
+  missScarlet,
+  mrsPeacock,
+  colonelMustard,
+  mrsGrey,
```
AxelHammarback, line 162: A bit of nitpicking, but you can omit the comma after the last item in an array (which I know you know since you did that for the weapons)

code/clue.js
```diff
@@ -1,59 +1,304 @@
...
+// CREATE AN OBJECT THAT KEEPS THE MYSTERY.
+// With a killer, a weapon and a room.
+// The values will be set later.
 
-const rooms = []
+const mystery = {
+  killer: {},
+  weapon: {},
+  room: {}
+};
```
AxelHammarback: If "fewer lines of code" is a pursuit, this could be replaced with:
`const mystery = {};`
Since assigning properties to an empty object creates both the key and the value. But I do like this approach due to the foresighted-ness (probably not a word) it leads to (i.e. me being able to predict what the object will be about).

code/clue.js
```diff
@@ -1,59 +1,304 @@
...
 
-// CREATE AN OBJECT THAT KEEPS THE MYSTERY.
-// With a killer, a weapon and a room.
-// The values will be set later.
+//----- Killer card -----//
+//Function is called when user clicks on card. The animation is run for 1.5 seconds, which delays the showPickKiller function from being called for that duration. Once over it will then call the showPickKiller function and the code inside of it e.g. show the card-details for that killer. The animation will be done by adding the animated-loader styling to the loader1 id.
+const pickKillerLoader = () => {
+  document.getElementById("loader1").classList.add("animated-loader");
+  setTimeout(showPickKiller, 2000);
+};
```
AxelHammarback: Very nice way to go about with the loading animation – today, I learned this ☝️

code/clue.js
```diff
@@ -1,59 +1,304 @@
...
+const showPickKiller = () => {
+  mystery.killer = randomSelector(suspects);// Will randomly select a killer from the suspects + add that to the mystery object.
+  document.getElementById("loader1").style.display = "none";//Will hide the loader image
+
+  document.getElementById('killerCard').removeEventListener('click', pickKillerLoader); //Will stop the loader showing next time you click the card
+
+  // Will change the background color of the card, the first/last name, the image and descrption for the selected killer. 
+  document.getElementById("killerCard").style.background = mystery.killer.color;
+  document.getElementById("killerName").innerHTML = `${mystery.killer.name}`;
+  document.getElementById("killerImage").src = mystery.killer.image;
```
AxelHammarback: I see some opportunities for optimisation here, which is to – at the start of this function* – assign variables to the duplicated `document.getElementById("X")` parts.

* Maybe even at the start of the javascript file – since it looks like these variables is used throughout multiple functions. (To the code coach who'll be reviewing my review; does it sounds like that would make sense?)

AxelHammarback: (And, another nit-picking note (sorry!): for consistency, try to avoid mixing ' and ")

code/clue.js
```diff
@@ -1,59 +1,304 @@
...
+  document.getElementById("roomLastSeen").innerHTML = `Last seen there: ${mystery.room.lastseen}`;
+  document.getElementById("roomfave").innerHTML = `Favourite room: ${mystery.room.favespot}`;
+
+  document.getElementById("cardDetails3").style.display = "flex";
+};
+
+//----- Function for on button click reveal mystery or get a message asking you to choose each card to then beable to reveal the mystery-----//
+function revealMystery() {
+  if (mystery.killer.hasOwnProperty("name") === true) {
+    document.getElementById("mystery").innerHTML = `The murder was commited by ${mystery.killer.name} with the ${mystery.weapon.name.toLowerCase()} in the ${mystery.room.name.toLowerCase()}`;
```
AxelHammarback: One small issue here. If you (the user) only clicks the Killer card, and the presses "Reveal mystery", then this function will enter the "If" statement and try to set weapon.name and room.name to lowercase. The issue is that it will cause a console error, since neither the weapon nor room has been revealed yet.

Perhaps do a small re-write of the condition to say something in the lines of:
"If mystery killer AND room AND weapon has own property 'Name', THEN [do stuff].

code/index.html
```diff
@@ -5,32 +5,84 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <link rel="stylesheet" href="style.css" />
-    <title>Technigo Project Cluedo</title>
+    <link href="https://fonts.googleapis.com/css2?family=Open+Sans&family=Prata&display=swap" rel="stylesheet">
```
AxelHammarback, line 8: Well done with the use of font and the overall UI side of things. I really like the theme and aesthetics you went with!

code/index.html
```diff
@@ -5,32 +5,84 @@
...
-        <p>ROOM</p>
-        <p id="roomName"></p>
-      </article>
-    </section>
+    <section class="main-wrapper">
+      <header>
+        <div class="header-container">
+          <div class="header-title">
+            <h1>Cluedo</h1>
+            <img src="./assets/magnify-glass.png" alt="Image of a magnify glass" width = "90px;">
```
AxelHammarback: Awesome with the "alt" tag – it's a good habit to never omit it (and something I myself need to get better at).

code/index.html
```diff
@@ -5,32 +5,84 @@
...
+        <div class="card-header-container">
+          <h2>Click on a card to reveal a murderer, weapon and room, then reveal the crime!</h2>
+        </div>
+
+        <!-- KILLER CARD -->
+        <div class="card-container">
+          <article class="killer-deck" id="killerCard">
+            <div class="card-front" id="frontOfCard">
+              <h4 class="card-title">KILLER</h4>
+              <img  id="loader1" class="loader-image" src="assets/magnify-glass.png" height="90px";/>
```
AxelHammarback: Any reason why height="90px" is specified here in the  tag, and not in the .loader-image {} class in the css? I've heard it's not great practise to add styling directly in the HTML tag – but there might be a good reason for it.

code/style.css
```diff
@@ -1,58 +1,264 @@
+* {
+  box-sizing: border-box;
+}
+
 body {
   margin: 0;
-  font-family: "Courier New", Courier, monospace;
+  font-family: 'Open Sans', sans-serif;
+  font-family: 'Prata', serif;
```
AxelHammarback: Won't these two collide with each other? How will elements in the body know which of these two fonts to pick?

code/style.css
```diff
@@ -1,58 +1,264 @@
...
-  margin: 1px;
+  margin: 10px;
   font-weight: 300;
-  background-color: #fff;
+  border-radius: 10px;
+  padding: 10px;
+  font-size: 20px;
+  color: #000000;
+  font-family: 'Open Sans', sans-serif;
+}
```
AxelHammarback: If you find yourself adding the same properties for multiple selectors, one way could be to group them for several elements. In the above example, an alternative way to structure it could be:

```
h4, p {
  color: #000;
  font-family: 'Open Sans', sans-serif;
}

h4 {
  font-weight: 300;
  font-size: 23px;
}

p {
  margin: 10px;
  font-weight: 300;
  border-radius: 10px;
  padding: 10px;
  font-size: 20px;
}
```

Not a great example since it ended up using _more_ lines of code. But I find it's a zen-ish thing to do when I end up accidentally having many lines of duplicate properties scattered all over my CSS file.

code/style.css
```diff
@@ -1,58 +1,264 @@
...
+  border-radius: 5px;
+  background-image: url("./assets/diamond-background-card.png");
+  background-size: 20px;
+  cursor: pointer;
+  box-shadow: 0 3px 5px 0 #555252;
+}
+
+.room-deck p {
+  margin: 0px;
+}
```
AxelHammarback: Re: The above example I mentioned with extracting out shared properties – this is a perfect example, given the three classes room-deck, killer-deck and weapon-deck have a lot of identical properties. 

What I would do here is to:
i) take all the duplicate properties of these x-deck classes, 
ii) put them in a separate class (and call it maybe .deck), and 
iii) assign the HTML card article elements the additional class "deck", like follows:

```
.deck {
  width: 500px;
  height: 550px;
  border-radius: 5px;
  cursor: pointer;
  box-shadow: 0 3px 5px 0 #555252;
  background-image: url("./assets/diamond-background-card.png");
  background-size: 20px;
}

.deck p {
  margin: 0px;
}

```
...and in the HTML file, you could structure the elements as such:

```
<article class="deck killer-deck" id="killerCard">...</article>
<article class="deck weapon-deck" id="weaponCard">...</article>
<article class="deck room-deck" id="roomCard">...</article>
```
Adding multiple classes to a HTML element by separating them with a blank space is (in my opinion) a great way to get some peace of mind. But yeah, totally up to you :)

code/style.css
```diff
@@ -1,58 +1,264 @@
...
+  display: block;
+  justify-content: center;
+  align-items: center;
+  margin-top: 30px;
+}
+
+.card-front img {
+  display: block;
+  margin-top: 50px;
+  margin-left: 165px;
```
AxelHammarback, line 124: On my monitor (wide-screen), I noticed that the mag glass icon wasn't centered. If you replace the "margin-left: 165px" to
```
margin-left: auto; 
margin-right: auto;
```
everything looked a-ok.

code/style.css
```diff
@@ -1,58 +1,264 @@
...
+}
+
+button:hover {
+  background: #ffffff;
+  color: #000000;
+  cursor: pointer;
 }
+
+.reveal-mystery-container {
+  background: #cccccc;
```
AxelHammarback: Final note; if you have a six-digit hex code which consists of identical characters, you can cut it in half. 

I.e. "#000000" is exactly the same as "#000", and "#cccccc" is exactly the same as "#ccc".

Not a major thing, but seeing as how it looks like you'll have a long and successful career of front-end development ahead of you – it can potentially save you MINUTES in the long run 🙂

AxelHammarback (commented):
Claire, you're a superstar! ⭐ 

A few notes on the overall PR – see more detailed notes and feedback in the code below.

* Awesomely done on documenting your process, the problem, and learnings. Reading through your reasoning, as well as code, you can really tell that you’re doing this with a lot of curiosity and excitement, which really is inspiring to the reader. 
* I really like how you structured your JS code into sections pertaining to the various areas – it made it easy to read and follow. 
* Well done with camelCase/kebab-case consistency as well as naming variables and functions in a logical manner.
* Excellently commented throughout – made this review a breeze.

Also, final note – even though I've made a few suggestions on things _I_ would do to improve this code, it's of course all up to you.

README.md
EmOlsz: Truly! :)

code/clue.js
EmOlsz: Actually, it's considered a good practice to add comma on the last property of object - it's because if you need to add new property, you are 100% sure that you don't have to worry about adding comma :)

code/clue.js
EmOlsz: The best would be 
```
const mystery = {
  killer: null,
  weapon: null,
  room: null
}
```

code/clue.js
EmOlsz: 👍

code/clue.js
EmOlsz: **_(To the code coach who'll be reviewing my review; does it sounds like that would make sense?)_** - Yes Axel, that do make sense, good assumption :)

**_(And, another nit-picking note (sorry!): for consistency, try to avoid mixing ' and ")_** Agree as well!

code/clue.js
```diff
@@ -1,59 +1,304 @@
...
+  mystery.killer = randomSelector(suspects);// Will randomly select a killer from the suspects + add that to the mystery object.
+  document.getElementById("loader1").style.display = "none";//Will hide the loader image
+
+  document.getElementById('killerCard').removeEventListener('click', pickKillerLoader); //Will stop the loader showing next time you click the card
+
+  // Will change the background color of the card, the first/last name, the image and descrption for the selected killer. 
+  document.getElementById("killerCard").style.background = mystery.killer.color;
+  document.getElementById("killerName").innerHTML = `${mystery.killer.name}`;
+  document.getElementById("killerImage").src = mystery.killer.image;
+  document.getElementById("killerDescription").innerHTML = `${mystery.killer.description}`;
```
EmOlsz: If there is no other string over here, you could just write
```
  document.getElementById("killerDescription").innerHTML = mystery.killer.description;
```

code/clue.js
```diff
@@ -1,59 +1,304 @@
...
+  document.getElementById("roomImage").src = mystery.room.image;
+  document.getElementById("roomLastSeen").innerHTML = `Last seen there: ${mystery.room.lastseen}`;
+  document.getElementById("roomfave").innerHTML = `Favourite room: ${mystery.room.favespot}`;
+
+  document.getElementById("cardDetails3").style.display = "flex";
+};
+
+//----- Function for on button click reveal mystery or get a message asking you to choose each card to then beable to reveal the mystery-----//
+function revealMystery() {
+  if (mystery.killer.hasOwnProperty("name") === true) {
```
EmOlsz: This is fine enough :

```
if (mystery.killer.name)
```

code/clue.js
EmOlsz: Great comment Axel, good job

code/clue.js
```diff
@@ -1,59 +1,304 @@
...
 
-// This function will be invoked when you click on the killer card.
-const pickKiller = () => {
-  // This will randomly select a killer from the suspects. And add that to the mystery object.
-  mystery.killer = randomSelector(suspects)
+//----- Shuffle function -----//
+// Allows for the weapons to be shuffled between the suspects
+const shuffleFavouriteWeapon = () => {
+  let copyWeapons = weapons.slice(0); //Creates shallow copy of weapons array
+  for (let i = 0; i < suspects.length; i++ ) {
```
EmOlsz, line 298: very nice Claire :)

code/clue.js
```diff
@@ -1,59 +1,304 @@
...
 
-  // This will change the background color of the card to the one connected to the chosen killer and show the full name of the killer. Feel free to add more things to show about the killer.
-  document.getElementById('killerCard').style.background = mystery.killer.color
-  document.getElementById(
-    'killerName'
-  ).innerHTML = `${mystery.killer.firstName} ${mystery.killer.lastName}`
-}
+//Function for re-loading the page
+const resetPage = () => {
+  location.reload();
```
EmOlsz: Hmm, wouldn't it be better to just clear mystery object and selected cards?

code/index.html
```diff
@@ -5,32 +5,84 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <link rel="stylesheet" href="style.css" />
```
EmOlsz: `./`

code/index.html
EmOlsz: Exactly, these styles should be stored in CSS file

code/style.css
EmOlsz: 👍

code/style.css
```diff
@@ -1,58 +1,264 @@
...
 
-.cards {
-  margin: 50px auto;
+.cards {  
+  background: #cccccc;
+  margin: 30px auto;
   width: 90%;
-  height: 600px;
-  background: #ccc;
+  /* height: 900px; */
```
EmOlsz: commented code should be removed before `git push`

code/style.css
```diff
@@ -1,58 +1,264 @@
...
+  margin-right: 30px;
+  border-radius: 5px;
+  background-image: url("./assets/diamond-background-card.png");
+  background-size: 20px;
+  cursor: pointer;
+  box-shadow: 0 3px 5px 0 #555252;
+}
+
+.killer-deck p {
+  margin: 0px;
```
EmOlsz: No need for `px`, if value is `0` - `margin: 0;` is completely enough :)

code/style.css
EmOlsz: Good job Axel!

code/style.css
```diff
@@ -1,58 +1,264 @@
...
+button:hover {
+  background: #ffffff;
+  color: #000000;
+  cursor: pointer;
 }
+
+.reveal-mystery-container {
+  background: #cccccc;
+  width: 60%;
+  margin: 0px auto;
```
EmOlsz: `0 auto`

EmOlsz (changes requested):
Great job Claire!:) I want to congratulate both of you - project and review are brilliant and very detailed. Moreover, I want to say that I see a lot of progress and advanced techniques in your code Claire, I'm more than happy to see that!
Before final approval, let's add fixes to comments that both Axel and I have created!

code/style.css
ClaireCaudwell: I thought that you are meant to specify all the font families I want to use in the body. Then I can specify which one I want to use in the specific class names? But ok I understand what you mean. So maybe it's better to only specify the font-family within the specific css class, rather than how I've done? @EmOlsz some clarification on how I should do this would be great. Thanks!

code/style.css
ClaireCaudwell: Very true Axel! I will implement this idea for my future projects as it makes more sense to have one class name for elements that are being styled the same way 👍

code/style.css
ClaireCaudwell: Amzing advice! Of course so simple. Will implement this!

code/style.css
ClaireCaudwell: Totally make sense. Thank you @EmOlsz and Axel! I just want to say that I think the quality of notes from Axel were really clear and thorough. So thank you!

code/style.css
EmOlsz: No, we don't have to specify all font families we are going to use, only the one we want to use.
Moreover, you don't need to add font families to `<body />` only, you can add them to any HTML element

code/clue.js
ClaireCaudwell: @EmOlsz I changed the mystery object properties to null. So when I went to change what you suggested in the if statement I got an error message "Cannot read property 'name' of null at revealMystery". So changed it to "mystery.killer != null" instead so it would work inline with the mystery object properties being set to null. And if the user clicks on the first card it won't be null anymore, otherwise if they try to revel the mystery before clicking on the cards then they will get the message from the else statement. I hope it's ok to write this instead of what you suggested?

code/clue.js
ClaireCaudwell: @EmOlsz How would I go about doing this? Sorry I'm uncertain of how to do this.

code/style.css
ClaireCaudwell: @EmOlsz I tried changing these margins to auto like Axel suggested, but it ended up shifting my magnify glass on each card out of the center where I had positioned them by using the specific pixels. Is there any way that you can use a standard,, like Axel suggested making them fix to the center of the card when being viewed on different screen sizes? Without having to adjust the margins for each screen size?

code/clue.js
ClaireCaudwell: I implemented what Axel suggested here and it returned no errors. 
```
function revealMystery() {
  if (mystery.killer != null && mystery.weapon != null && mystery.room != null )
```
Unlike before when I had only specified mystery.killer. Such a good tip and another factor to remember to consider when testing out onclick events to and see what errors pop up.

code/clue.js
ClaireCaudwell: Thank you @EmOlsz but I cannot take all the credit. Got some help from my partner on this one!

code/clue.js
ClaireCaudwell: @AxelHammarback Thanks for this protip Axel. This was a really good learning moment for me!

code/index.html
ClaireCaudwell: Thanks @AxelHammarback

code/style.css
ClaireCaudwell: @AxelHammarback I decided to implement this as I thought it would make my code cleaner! Thanks for the pro-tip.

code/clue.js
AxelHammarback: Good to know – thanks, Maks!

EmOlsz (approved):

PR #112, opened 2020-09-18
Cluedo project Sandra Shumanteva Kaevska

sandra-shumanteva-kaevska:
I have done the blue, red and black level for this project.

### DIFF
--- a/code/clue.js
+++ b/code/clue.js
@@ -1,59 +1,221 @@
 // STEP 1 - CREATE OBJECTS FOR ALL THE SUSPECTS, SOMETHING LIKE THIS:
 
 const mrGreen = {
-  firstName: 'Jacob',
-  lastName: 'Green',
-  color: 'green',
-  description: 'He has a lot of connections',
+  firstName: "Jacob",
+  lastName: "Green",
+  color: "green",
+  description: "He has a lot of connections",
   age: 45,
-  image: 'assets/green.png',
-  occupation: 'Entrepreneur'
-}
+  image: "assets/green.png",
+  occupation: "Entrepreneur",
+};
+
+const professorPlum = {
+  firstName: "Victor",
+  lastName: "Plum",
+  color: "purple",
+  description: "He has a lot of medicin knowledge",
+  age: 40,
+  image: "assets/plum.png",
+  occupation: "professor",
+};
+
+const missScarlet = {
+  firstName: "Cassandra",
+  lastName: "Scarlet",
+  color: "red",
+  description: "She is creative",
+  age: 35,
+  image: "assets/scarlet.png",
+  occupation: "architect",
+};
+
+const mrsPeackok = {
+  firstName: "Eleanor",
+  lastName: "Peacock",
+  color: "blue",
+  description: "She is free-minded",
+  age: 55,
+  image: "assets/peacock.png",
+  occupation: "actress",
+};
+
+const colonelMustard = {
+  firstName: "Jack",
+  lastName: "Mustard",
+  color: "yellow",
+  description: "He is strong",
+  age: 34,
+  image: "assets/mustard.png",
+  occupation: "butcher",
+};
+
+const mrsWhite = {
+  firstName: "Ana",
+  lastName: "White",
+  color: "white",
+  description: "She is very weak",
+  age: 28,
+  image: "assets/white.png",
+  occupation: "painter",
+};
 
 // CREATE OBJECTS FOR ALL THE WEAPONS, ADD MORE CHARACTERISTICS TO THE WEAPONS IF YOU LIKE.
 
 const rope = {
-  name: 'rope',
-  weight: 10
-}
+  name: "rope",
+  weight: 10,
+};
+
+const knife = {
+  name: "knife",
+  weight: 5,
+};
+
+const candlestick = {
+  name: "candelstick",
+  weight: 7,
+};
+
+const dumbbell = {
+  name: "dumbell",
+  weight: 30,
+};
+
+const poison = {
+  name: "dumbell",
+  weight: 2,
+};
+
+const axe = {
+  name: "axe",
+  weight: 12,
+};
+
+const bat = {
+  name: "bat",
+  weight: 13,
+};
+
+const trophy = {
+  name: "trophy",
+  weight: 20,
+};
+
+const pistol = {
+  name: "pistol",
+  weight: 90,
+};
 
 // THE ROOMS ONLY HAS A NAME SO NO NEED FOR OBJECTS THERE.
+
+console.log(colonelMustard.color);
+console.log(pistol.weight);
 
 // NOW GROUP ALL SUSPECTS, WEAPONS AND ROOMS IN ARRAYS LIKE THIS:
 
 const suspects = [
   mrGreen,
-  mrsWhite
-  // ...  and the rest
-]
+  professorPlum,
+  missScarlet,
+  mrsPeackok,
+  colonelMustard,
+  mrsWhite,
+];
 
-const weapons = []
+const weapons = [
+  rope,
+  knife,
+  candlestick,
+  dumbbell,
+  poison,
+  axe,
+  bat,
+  trophy,
+  pistol,
+];
 
-const rooms = []
+const rooms = [
+  "Dining Room",
+  "Conservatory",
+  "Kitchen",
+  "Study",
+  "Library",
+  "Billiard Room",
+  "Lounge",
+  "Ballroom",
+  "Hall",
+  "Spa",
+  "Living Room",
+  "Observatory",
+  "Theater",
+  "Guest House",
+  "Patio",
+];
+
+console.log(suspects);
+console.log(suspects[3].occupation);
+console.log(weapons);
+console.log(rooms.indexOf("Spa"));
 
 // THIS FUNCTION WILL RANDOMLY SELECT ONE ITEM FROM THE ARRAY THAT YOU PASS IN TO THE FUNCTION.
 // YOU DON'T NEED TO CHANGE THIS, JUST TRY TO UNDERSTAND IT. AND HOW TO USE IT.
-const randomSelector = array => {
-  return array[Math.floor(Math.random() * array.length)]
-}
+
+const randomSelector = (array) => {
+  return array[Math.floor(Math.random() * array.length)];
+};
 
 // CREATE AN OBJECT THAT KEEPS THE MYSTERY.
 // With a killer, a weapon and a room.
 // The values will be set later.
+const mystery = {
+  killer: undefined,
+  weapon: undefined,
+  room: undefined,
+};
 
 // This function will be invoked when you click on the killer card.
 const pickKiller = () => {
   // This will randomly select a killer from the suspects. And add that to the mystery object.
-  mystery.killer = randomSelector(suspects)
+  mystery.killer = randomSelector(suspects);
 
   // This will change the background color of the card to the one connected to the chosen killer and show the full name of the killer. Feel free to add more things to show about the killer.
-  document.getElementById('killerCard').style.background = mystery.killer.color
+
+  document.getElementById("killerCard").style.background = mystery.killer.color;
   document.getElementById(
-    'killerName'
-  ).innerHTML = `${mystery.killer.firstName} ${mystery.killer.lastName}`
-}
+    "killerName"
+  ).innerHTML = `${mystery.killer.firstName} ${mystery.killer.lastName}`;
+};
 
 // CREATE FUNCTIONS pickWeapon and pickRoom in a similar way.
 
+const pickWeapon = () => {
+  mystery.weapon = randomSelector(weapons);
+  document.getElementById(
+    "weaponCard"
+  ).innerHTML = `${mystery.weapon.name} which weight is ${mystery.weapon.weight}.`;
+};
+
+const pickRoom = () => {
+  mystery.room = randomSelector(rooms);
+  document.getElementById("roomCard").innerHTML = mystery.room;
+};
+
+// Invoke functions
+
+document.getElementById("killerCard").onclick = pickKiller;
+
+document.getElementById("weaponCard").onclick = pickWeapon;
+
+document.getElementById("roomCard").onclick = pickRoom;
+
 // STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
 // 'The murder was committed by Jacob Green, in the living room with a rope.'
+
+const revalMystery = () => {
+  document.getElementById(
+    "reveal"
+  ).innerHTML = `The murder was committed by ${mystery.killer.firstName} ${mystery.killer.lastName}, in the ${mystery.room} with a ${mystery.weapon.name}.`;
+};
+
+document.getElementById("reveal").onclick = revalMystery;
--- a/code/clue.js
+++ b/code/clue.js
@@ -8,7 +8,7 @@ const mrGreen = {
   age: 45,
   image: "assets/green.png",
   occupation: "Entrepreneur",
-  favouriteWeapon: "knife",
+  favouriteWeapon: null,
 };
 
 const professorPlum = {
@@ -19,7 +19,7 @@ const professorPlum = {
   age: 40,
   image: "assets/plum.png",
   occupation: "professor",
-  favouriteWeapon: "rope",
+  favouriteWeapon: null,
 };
 
 const missScarlet = {
@@ -30,7 +30,7 @@ const missScarlet = {
   age: 35,
   image: "assets/scarlet.png",
   occupation: "architect",
-  favouriteWeapon: "candlestick",
+  favouriteWeapon: null,
 };
 
 const mrsPeackok = {
@@ -41,7 +41,7 @@ const mrsPeackok = {
   age: 55,
   image: "assets/peacock.png",
   occupation: "actress",
-  favouriteWeapon: "dumbbell",
+  favouriteWeapon: null,
 };
 
 const colonelMustard = {
@@ -52,7 +52,7 @@ const colonelMustard = {
   age: 34,
   image: "assets/mustard.png",
   occupation: "butcher",
-  favouriteWeapon: "poison",
+  favouriteWeapon: null,
 };
 
 const mrsWhite = {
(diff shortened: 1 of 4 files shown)


code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+  document.getElementById(
+    "mysteryName"
+  ).innerHTML = `Favorite wepon is ${mystery.favWeapon.id}`;
+};
+
+const shuffle = (favArr, arr) => {
+  const favDuplicated = duplicateItems(favArr);
+  const concatArr = [...arr, ...favDuplicated];
+  return randomSelector(concatArr);
+};
```
HenrikeW: Great that you managed to do a function to shuffle the favourite weapons! I had to think about it for a while but I think it does something different from what was asked in the requirements. I don't know if we understood the requirements differently and your function does exactly what you want it to do, or if this is unintended and your functions does something else than what you want it to do. 

How I understood the requirements:
- every suspect should have a favourite weapon
- this favourite weapon should randomly be assigned to the suspect by a shuffleFavouriteWeapon-function
- in the next step, after a killer is chosen, this person's favourite weapon should be more likely to be picked when choosing a weapon. 

What (I think) your code does:
- it gives every suspect a favourite weapon
- it groups these 6 favourite weapons into an array
- it creates a new array that contains these favourite weapons multiple times plus each of the original 9 weapons one time
- it chooses a weapon from this new array

As a result, it becomes more likely to choose **someone's** favourite weapon, but not the favourite weapon of a specific person. 

In case this is not what you intended to do, here are some suggestions for changes: 

To give the killer a random favourite weapon, you could just use the randomSelector to create a function:
`const shuffleFavouriteWeapon = person => person.favouriteWeapon = randomSelector(weapons).id;`
and then call it inside the pickKiller-function with the chosen killer as an argument:
`mystery.killer = randomSelector(suspects);
  shuffleFavouriteWeapon(mystery.killer);`

Then you could create a new array with the original weapons and several copies of this favourite weapon and use that array to pick a weapon for the weapon deck.

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
 // With a killer, a weapon and a room.
 // The values will be set later.
+const mystery = {
+  killer: undefined,
+  weapon: undefined,
+  room: undefined,
+  favWeapon: undefined,
+};
+
+// shuffleFavouriteWeapon function
```
HenrikeW: After this comment I got confused when the next function I saw was called "duplicate items". Maybe you could have made the comment a bit longer, something like "shuffleFavouriteWeapon function with supporting functions duplicateItems and shuffle" and changed the order of the following functions.
Most intuitive for me would have been this:
1. duplicateItems
2. shuffle
3. shuffleFavouriteWeapon

or this:
1. shuffleFavouriteWeapon
2. shuffle
3. duplicateItems

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+};
+
+// LOADER
+
+function loader() {
+  document.getElementById("loaderItem").style.display = "block";
+}
+
+// Defaoult function
+const reset = () => {
```
HenrikeW, line 337: Right now, there are some issues with the loader: 
- when I click on Reset before clicking on "Reveal the crime" the loader stays which confuses me as a user (like, "what is loading right now??")
- when have revealed a crime, but not clicked Reset, and then click on a card deck again, the loader reappears and there is no way to make it disappear, except clicking all the four card decks and then "Reveal the crime".

Including one line in the reset function could solve the issue: 
`  document.getElementById("loaderItem").style.display = "none";`

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+  document.getElementById("killerCard").style.background = "";
+  document.getElementById("killerName").innerHTML = "";
+  document.getElementById("killerOccupation").innerHTML = "";
+  document.getElementById("killerImage").innerHTML = "";
+  document.getElementById("weaponName").innerHTML = "";
+  document.getElementById("roomName").innerHTML = "";
+  document.getElementById("reveal").innerHTML = "REVEAL THE CRIME";
+};
+
+const deleteValue = () => {};
```
HenrikeW: I don't understand what this line does. Is it some kind of leftover?

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+  pistol: ["Living Room", "Gest House"],
+};
 
 // THIS FUNCTION WILL RANDOMLY SELECT ONE ITEM FROM THE ARRAY THAT YOU PASS IN TO THE FUNCTION.
 // YOU DON'T NEED TO CHANGE THIS, JUST TRY TO UNDERSTAND IT. AND HOW TO USE IT.
-const randomSelector = array => {
-  return array[Math.floor(Math.random() * array.length)]
-}
+
+const randomSelector = (array) => {
```
HenrikeW, line 202: When there is only one argument in the function, there is a convention not to use parenthesis around the argument, in this case `array`. Like this: 

`const randomSelector = array => {  
return array[Math.floor(Math.random() * array.length)];
};`

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+];
 
-const weapons = []
+console.log(suspects);
+console.log(suspects[3].occupation);
+console.log(weapons);
+console.log(rooms.indexOf("Spa"));
 
-const rooms = []
+const weaponRooms = {
```
HenrikeW, line 167: Such a good idea to solve the problem this way, to make certain weapons more likely in certain rooms. I like that you can be so flexible, having e.g. only one room for the knife, but many for the candlestick!

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+  "Theater",
+  "Guest House",
+  "Patio",
+];
 
-const weapons = []
+console.log(suspects);
+console.log(suspects[3].occupation);
+console.log(weapons);
+console.log(rooms.indexOf("Spa"));
```
HenrikeW: Same as with the console logs above...

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+const pistol = {
+  name: "pistol",
+  weight: 90,
+  id: "pistol",
+};
 
 // THE ROOMS ONLY HAS A NAME SO NO NEED FOR OBJECTS THERE.
 
+console.log(colonelMustard.color);
+console.log(pistol.weight);
```
HenrikeW: These console logs are there for you, to check if your code works, right? In that case, you should delete them before handing in the project, I think.

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+};
+
+const colonelMustard = {
+  firstName: "Jack",
+  lastName: "Mustard",
+  color: "yellow",
+  description: "He is strong",
+  age: 34,
+  image: "assets/mustard.png",
+  occupation: "butcher",
```
HenrikeW, line 54: I love that he's a butcher :D

code/clue.js
```diff
@@ -1,59 +1,350 @@
 // STEP 1 - CREATE OBJECTS FOR ALL THE SUSPECTS, SOMETHING LIKE THIS:
 
 const mrGreen = {
-  firstName: 'Jacob',
-  lastName: 'Green',
-  color: 'green',
-  description: 'He has a lot of connections',
+  firstName: "Jacob",
```
HenrikeW, line 4: So good to be consistent with the quotation marks! 👍

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+
+const mrsPeackok = {
+  firstName: "Eleanor",
+  lastName: "Peacock",
+  color: "blue",
+  description: "She is free-minded",
+  age: 55,
+  image: "assets/peacock.png",
+  occupation: "actress",
+  favouriteWeapon: "dumbbell",
```
HenrikeW: Nice favourite weapon ;-)

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+const shuffleFavRoom = () => {
+  if (mystery.weapon !== undefined) {
+    const roomList = weaponRooms[mystery.weapon.id];
+    mystery.room = shuffle(roomList, rooms);
+    document.getElementById("roomName").innerHTML = mystery.room;
+  } else {
+    mystery.room = randomSelector(rooms);
+    document.getElementById("roomName").innerHTML = mystery.room;
+  }
+};
```
HenrikeW: This seems to be a really smooth way of achieving the goal, having a certain weapon more likely to a appear in certain rooms. Great job! 💯

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+  document.getElementById("killerName").innerHTML =
+    mystery.killer.firstName + " " + mystery.killer.lastName;
+
+  document.getElementById("killerOccupation").innerHTML =
+    mystery.killer.occupation;
+
   document.getElementById(
-    'killerName'
-  ).innerHTML = `${mystery.killer.firstName} ${mystery.killer.lastName}`
-}
```
HenrikeW: Is there a reason that you deleted the killerName-part with template literals and chose the string + string version in line 270 instead? I think, template literals are the way to go.

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
 // CREATE FUNCTIONS pickWeapon and pickRoom in a similar way.
 
+const pickWeapon = () => {
+  mystery.weapon = randomSelector(weapons);
+  document.getElementById(
+    "weaponName"
+  ).innerHTML = `${mystery.weapon.name} which weight is ${mystery.weapon.weight}.`;
+};
+
+// LOADER
```
HenrikeW, line 294: Great that you added a loader!

code/clue.js
HenrikeW, line 337: In general: Very good idea to include a reset button even if it wasn't named in the requirements. That really improves the user experience. 👍

code/style.css
```diff
@@ -41,18 +41,69 @@ p {
...
+    -webkit-transform: rotate(360deg);
+  }
+}
+
+@keyframes spin {
+  0% {
+    transform: rotate(0deg);
+  }
+  100% {
+    transform: rotate(360deg);
```
HenrikeW: Nice that you have an animated loader!

code/style.css
```diff
@@ -41,18 +41,69 @@ p {
...
+  background: rgb(112, 25, 86);
   text-align: center;
   padding-top: 30px;
   border-radius: 16px;
 }
 
 button {
   padding: 8px 16px;
-  background: crimson;
+  background: greenyellow;
```
HenrikeW, line 60: Smart to make the button green, in contrast to the red Reset-button! It's a detail, but it's really good that you thought about it. 👍

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+  shuffleFavRoom();
+};
+
+document.getElementById("mysteryCard").onclick = shuffleFavouriteWeapon;
+
 // STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
-// 'The murder was committed by Jacob Green, in the living room with a rope.'
+// 'The murder was committed by Jacob Green, in the living room with a rope. '
+
+const revalMystery = () => {
```
HenrikeW, line 326: Just as an idea: It could be nice to include a message to the user if not all of the four decks have been clicked, something like "Please choose one card from each deck". Otherwise, the user could start to wonder if the "Reveal the mystery" button is working properly.

HenrikeW (approved):
I think you did a really good job, achieving not only the blue but also the red and the black level. Especially your solution for weapons being more likely in certain rooms is both flexible and smooth at the same time. Congratulations :)
I'm not sure if your shuffleFavouriteWeapon-function does what it is supposed to do, but you'll find a more detailed comment about that in the code. 
Your code is easy to follow (most of the times), you're consistent with quotation marks and semi colons and the indentation is good. Continue this way :)

code/clue.js
EmOlsz: Good job Henrike!

code/clue.js
EmOlsz: 👍

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+};
 
 // CREATE AN OBJECT THAT KEEPS THE MYSTERY.
 // With a killer, a weapon and a room.
 // The values will be set later.
+const mystery = {
+  killer: undefined,
+  weapon: undefined,
+  room: undefined,
+  favWeapon: undefined,
```
EmOlsz: It's far better to initialise your object with `null` values

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+
+const shuffle = (favArr, arr) => {
+  const favDuplicated = duplicateItems(favArr);
+  const concatArr = [...arr, ...favDuplicated];
+  return randomSelector(concatArr);
+};
+// ----------------WEAPON AND ROOM
+
+const shuffleFavRoom = () => {
+  if (mystery.weapon !== undefined) {
```
EmOlsz: `if (!mystery.weapon) { ... }` is another way to do it

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+};
+
+// LOADER
+
+document.getElementById("loaderItem").style.display = "none";
+
+// Invoke function
+
+document.getElementById("killerCard").onclick = () => {
+  loader();
```
EmOlsz, line 306: It's a bit confusing that even after you pick single card (suspect, weapon or room), spinner is still spinning - it could be stopped until next selection

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+  pickKiller();
+};
+
+document.getElementById("weaponCard").onclick = () => {
+  loader();
+  pickWeapon();
+};
+document.getElementById("roomCard").onclick = () => {
+  loader();
+  // pickRoom();
```
EmOlsz: commented code

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+
+const revalMystery = () => {
+  document.getElementById(
+    "reveal"
+  ).innerHTML = `The murder was committed by ${mystery.killer.firstName} ${mystery.killer.lastName}, in the ${mystery.room} with a ${mystery.weapon.name}. Favorite wapon is ${mystery.favWeapon.id}`;
+};
+
+document.getElementById("reveal").onclick = () => {
+  revalMystery();
+  document.getElementById("loaderItem").style.display = "none";
```
EmOlsz: This code could be moved to separate function

code/clue.js
EmOlsz: Good point

code/clue.js
```diff
@@ -1,59 +1,350 @@
...
+function loader() {
+  document.getElementById("loaderItem").style.display = "block";
+}
+
+// Defaoult function
+const reset = () => {
+  delete mystery.killer;
+  delete mystery.weapon;
+  delete mystery.room;
+  delete mystery.favWeapon;
```
EmOlsz: I would rather set mystery like this :
```
mystery = {}
```

it's far more clean approach, we should avoid `delete` keyword. It will be more understandable fo you when we will jump into React

code/style.css
```diff
@@ -41,18 +41,69 @@ p {
...
+    transform: rotate(0deg);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+}
+
+/* RESET BTN */
+
+#resetBtn {
```
EmOlsz: We should avoid styling by id's, let's use class instead

EmOlsz (changes requested):
Good job Sandra! Your project is pretty complex, I like that you've added more than blue requirements :)
Nonetheless, I see some leftovers here and there - please, refer to both Henrike's and my comments and fix your code. Only then I will be able to approve your application.
For the future, try to double check everything before final push, messy code is as bad as bugs!

code/clue.js
sandra-shumanteva-kaevska: Good catch, I have missed that part. Now the logic is updated 👍

code/clue.js
sandra-shumanteva-kaevska: I did some rethinking here and renamed the functions

code/clue.js
sandra-shumanteva-kaevska: I updated the loader to work according the documentations. Now it stays active for a second after some card is clicked

code/clue.js
sandra-shumanteva-kaevska: Removed

code/clue.js
sandra-shumanteva-kaevska: As I can read from the JS documentations, the { brackets } and ( parentheses ) and "return" are optional, but may be required. Can you please share where you have read that convention not to use parenthesis around one argument?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

Also my VSCode by default formats the JS by adding those parentheses if I try to remove them

code/clue.js
sandra-shumanteva-kaevska: Updated

code/clue.js
sandra-shumanteva-kaevska: I updated them to null, but as I remember on one of the lessons in similar scenario it was mentioned that values should be undefined.
Can you please share why its null better than undefined in this case?

code/clue.js
sandra-shumanteva-kaevska: Done

sandra-shumanteva-kaevska:
I have done the update that you have suggested and also have left some comments that I would like to get more info about.

Thank you for the detailed and good review :)

code/clue.js
EmOlsz: Sandra, regarding your question, this is "longer" way of writing this precise function :

```
const randomSelector = array => {
  return array[Math.floor(Math.random() * array.length)];
};
```

And this is "shorter" version of the same function. By "shorter" I mean that we can write everything in one line if there is only one line inside "longer" version of function. When we create "shorter" version, we can omit two things : 1. brackets `{}` and 2. `return` keyword. JavaScript will add it by itself, so it's still in there, but "invisible" :

```
const randomSelector = array => array[Math.floor(Math.random() * array.length)];
```

code/clue.js
EmOlsz: Sure Sandra! We use `undefined` values as initial ones when we create variables that we will want to update later on. Good example of that can be for example this piece of code, which will show user on website different alert, depending on user age :

```
let greeting; // Over here we initialise variable as undefined. NOTE ONE THING : we don't have to write let greeting = undefined, because let greeting; will do the same
if (userAge > 10) {
    greeting = 'Hey champ';
    alert(greeting);
} else if (userAge > 25) {
    greeting = 'Hello there';
    alert(greeting);
} else if (userAge > 50) {
    greeting = 'Good morning sir';
    alert(greeting);
}
```

Null however we should use mostly when we are talking about properties in object, exactly as in `mystery` over here. So we already know that `mystery` type is object, we just don't know yet exact values of properties

EmOlsz (approved):

sandra-shumanteva-kaevska:
Thank you very much for the advice and addition explanation.

PR #121, opened 2020-09-20
Cluedo game - Hanna Skeppe

Hanna-Skeppe:
I would love to get comments on how to get the 'PLAY AGAIN'-button to reload the page, and just overall feedback on things to improve when it comes to js.

More info in the README.

For this week I completed the red level. I made animations on cards (and button) and shuffled weapons. Tried to make the random weapon not so random, but I didn't manage to make it work. 

https://suspicious-mahavira-031870.netlify.app/


code/index.html
```diff
@@ -4,34 +4,67 @@
...
     <title>Technigo Project Cluedo</title>
   </head>
   <body>
+   <section class="main-wrapper">
+    <header class="header">
+      <section class="header-content">
+        <h1 class="title-heading">
+          Cluedo
+        </h1>
+        <blockquote class="title-quote">
```
the-j-curl, line 22: I believe blockquote tag is usually used to signify an indented quote. Maybe a h2 or h3 tag may have been better suited?

code/index.html
```diff
@@ -4,34 +4,67 @@
...
-    </section>
+      <div class="card-container">
+        <article class="killer-deck" id="killerCard"> 
+          <!-- removed onclick="pickKiller()" above and added eventlistener in js below pickkiller function-->
+          <p class="card-title">KILLER</p>
+          <p id="killerName"></p>
+          <p id="killerAge"></p>
+          <p id="killerOccupation"></p>
+          <p id="killerWeapon"></p>
+          <img id="killerImage" src="" height="100" />
```
the-j-curl: The height attribute is inline CSS and as good practise could be moved to the CSS file.

code/index.html
```diff
@@ -4,34 +4,67 @@
...
+          <img id="weaponImage" src=""/>
+        </article>
+        <div class="loader" id="weaponLoader"></div>
+      </div>
 
-    <button>REVEAL THE CRIME</button>
+      <div class="card-container">
+        <article class="room-deck" id="roomCard">
+          <p class="card-title">ROOM</p>
+          <p id="roomName"></p>
```
the-j-curl, line 56: I have done exactly the same in my code, but as per Maks comments today maybe kebab case should be used for ID's going forward

code/style.css
```diff
@@ -1,58 +1,248 @@
 body {
   margin: 0;
-  font-family: "Courier New", Courier, monospace;
+}
+
+.main-wrapper {
+  background-image: url("assets/main-wrapper-background.jpg");
+  background-position: 55% 0%; /* Center the image */
```
the-j-curl, line 7: I really like the background image, it works very well with this weeks theme. It also looks like you have positioned it exactly where you want it, which is great.

code/style.css
```diff
@@ -1,58 +1,248 @@
...
+.header-content {
+  display: flex;
+  align-items: baseline;
+  justify-content: space-around;
+
+}
+
+.title-heading {
+  margin: 0;
+  font-family: 'IM Fell DW Pica SC', serif;
```
the-j-curl, line 22: This font has been well chosen, I like it!

code/style.css
```diff
@@ -1,58 +1,248 @@
...
 
 p {
+  font-family: 'Josefin Sans', sans-serif;
   margin: 1px;
   font-weight: 300;
   background-color: #fff;
+  line-height: 1.7;
 }
 
 .cards {
```
the-j-curl, line 41: I think the "cards" container helps to give your site a smart and elegant look.

code/index.html
```diff
@@ -4,34 +4,67 @@
...
+          <p id="weaponName"></p>
+          <p id="weaponWeight"></p>
+          <p class="weapon-flexibility" id="weaponFlexibility"></p>
+          <img id="weaponImage" src=""/>
+        </article>
+        <div class="loader" id="weaponLoader"></div>
+      </div>
 
-    <button>REVEAL THE CRIME</button>
+      <div class="card-container">
```
the-j-curl, line 53: I'd like to understand the purpose of having the extra card-container and why you added it?

code/style.css
```diff
@@ -1,58 +1,248 @@
...
   padding-top: 30px;
   border-radius: 16px;
 }
 
+.killer-deck {
+  background: #7a5948;
+  position: relative;
+}
+
+.killer-deck img,
```
the-j-curl, line 77: Good that you have styled these 3 classes together, it helps to keep the code clean.

code/style.css
```diff
@@ -1,58 +1,248 @@
...
+}
 
-  text-align: center;
-  padding-top: 30px;
-  border-radius: 16px;
+.room-image {
+  margin-top: 88px;
 }
 
+.cards article,
```
the-j-curl, line 106: I'm not sure what this is actually targeting? Could you have used the class name instead?

code/style.css
```diff
@@ -1,58 +1,248 @@
...
+  background: none;
+}
+
+.loader {
+  position: absolute;
+  top: 44%;
+  left: 40%;
+  width: 50px;
+  height: 50px;
+  /* background: #8a365d; */
```
the-j-curl: This commented out section could of been deleted.

code/style.css
```diff
@@ -1,58 +1,248 @@
...
+  background: #6e1411;
+  border-radius: 50%;
+  animation-duration: 2.5s;
+  animation-iteration-count: infinite;
+  animation-name: ripple1;
+  animation-timing-function: ease-out;
+  visibility: hidden;
+}
+
+@keyframes ripple1 {
```
the-j-curl, line 132: This seems like a complex animation and looks great. Nice work getting it all working.

code/style.css
```diff
@@ -1,58 +1,248 @@
...
-  margin: 0 auto;
+  margin: 10px auto;
+  text-align: center;
+  cursor: pointer;
+  animation: pulsate 4s ease-in;
+  animation-iteration-count: infinite;
+  outline: 0;
+}
+
+@keyframes pulsate {
```
the-j-curl, line 177: Is there supposed to be two animations? Pulsate and ripple are both listed.

code/style.css
```diff
@@ -1,58 +1,248 @@
...
+  }
+}
+
+@media(max-width:420px) {
+  .mystery-heading {
+    height: 180px;
+  }
+}
+
+@media(max-width:320px) {
```
the-j-curl, line 259: Great that you have added extra detail and made the site responsive.

code/clue.js
```diff
@@ -1,59 +1,396 @@
...
+  name: 'Guest House',
+  image: 'assets/guesthouse.jpg'
+};
+
+const patio = {
+  name: 'Patio',
+  image: 'assets/patio.jpg'
+};
+
+// STEP 2. NOW GROUP ALL SUSPECTS, WEAPONS AND ROOMS IN ARRAYS LIKE THIS:
```
the-j-curl: Really clean code and good that you have used the comments to explain more detail.

code/clue.js
```diff
@@ -1,59 +1,396 @@
...
+  guestHouse,
+  patio
+];
 
 // THIS FUNCTION WILL RANDOMLY SELECT ONE ITEM FROM THE ARRAY THAT YOU PASS IN TO THE FUNCTION.
 // YOU DON'T NEED TO CHANGE THIS, JUST TRY TO UNDERSTAND IT. AND HOW TO USE IT.
-const randomSelector = array => {
-  return array[Math.floor(Math.random() * array.length)]
+const randomSelector = array => { //I don't understand this
+  return array[Math.floor(Math.random() * array.length)] //What does array.length do here?
```
the-j-curl: This code generates a random index number. And array.length ensures that the number is relevant to the array in question. Math.random generates a random number between 0 - 1. If we were to multiply this by 10 then we would get a random number between 1 - 10. But by using this code it will multiply that random number by the length of the array. And means that the index number will be between 0 - 11 if the array in question has a length of 12. 
I hope this helps?

code/clue.js
```diff
@@ -1,59 +1,396 @@
...
+    }, 2500)
+  }
+};
+//Added eventlistener below instead of "onclick" in HTML. Don't know which way is the best way to do it. This only works if the eventlistener is outside of the function. 
+//First param: type of event. Second param: the function to call on event.
+document.getElementById('killerCard').addEventListener('click', pickKiller);
+
+//If pickKiller function is excecuted/card is clicked then excecute this function:
+const pickWeapon = () => {
+  if (killerIsPicked && !weaponIsPicked) {
```
the-j-curl, line 342: I like that you can only select a card once the previous animation has finished.

code/clue.js
```diff
@@ -1,59 +1,396 @@
...
+  //Invoke when button is clicked and show elements from mystery object in text:
+document.getElementById('mystery').innerHTML = `The murder was committed by ${mystery.killer.firstName} in the ${mystery.room.name} with ${mystery.weapon.name}` 
+document.getElementById('revealButton').innerHTML = `PLAY AGAIN?` //change the text on button.
 }
 
-// CREATE FUNCTIONS pickWeapon and pickRoom in a similar way.
+//BELOW I TRIED TO MAKE A FUNCTION SO THAT THE PAGE RELOADS WHEN 'PLAY AGAIN'-BUTTON IS CLICKED. IT DIDN'T WORK. I wanted to leave this here just to show the code-reviewer what I tried and maybe get pointers on how to achieve this.
+
+//THIS DOESN'T WORK:
+//  const playAgain = () => {
```
the-j-curl: The single line comments can look messy, try using /* */ to comment out multiple lines.

code/clue.js
```diff
@@ -1,59 +1,396 @@
...
+//    if (document.getElementById('revealButton').innerHtml === 'PLAY AGAIN?') {
+//         //location.reload();
+//         revealButton().onclick === pageReload; 
+//         //revealButton().onclick === location.reload();
+//     } 
+//   };
+//document.getElementById('revealButton').addEventListener('click', playAgain);
```
the-j-curl: I can't see exactly why this isn't working, but I have noticed that you haven't changed killerIsPicked, roomIsPicked, weaponIsPicked variables back to false.

the-j-curl (commented):
I really like your Cluedo site Hanna, it looks super cool! The whole design looks carefully thought out. I particularly like that you can only select one card at a time, after the animation has finished. 
In general your code looks very clean. I think using the multi-line comment in JavaScript will help going forward. 
Well done in completing the red level, I know the favourite weapon task wasn't easy!

code/clue.js
Hanna-Skeppe: good point 👍

code/clue.js
Hanna-Skeppe: Thank you, very clear explanation Jamie 👍

code/style.css
Hanna-Skeppe: Yes, you are right. I only use one of them so the other one I should have removed.

code/style.css
Hanna-Skeppe: 👍

code/index.html
Hanna-Skeppe: Yes, you're right.

code/style.css
puj: Agreed, quite cool and advanced. Nice job!

code/clue.js
```diff
@@ -1,59 +1,396 @@
...
+const mystery = {
+  killer: "",
+  weapon: "",
+  room: "",
+};
+
+//Function to shuffle the favourite weapons of the suspects in a randomized way.
+const shuffleWeapons = () => {
+  suspects.forEach(suspect => {
+    suspect.favorite = randomSelector(weapons).name
```
puj: An alternative here would be to 
`suspect.favorite = randomSelector(weapons)`

This would give you access to the entire weapon object that this suspect prefers, not just the name.  Could come in handy later, but your solution is precise👍

code/clue.js
puj: Indeed, nice touch!

code/clue.js
```diff
@@ -1,59 +1,396 @@
...
+
+//Function to increase probability of favourite weapon to be assigned to killer:
+//THIS FUNCTION DOESN'T WORK???
+const increaseChanceOfFavorite = () => {
+  const favWeapon = weapons.find(weapon => weapon.id === mystery.killer.favorite);
+  const favWeaponArray = new Array(7).fill(favWeapon);
+  console.log(favWeaponArray)
+  const calculatedWeapon = [...weapons, ...favWeaponArray];
+  return calculatedWeapon;
+};
```
puj: This code looks like it should work, but it requires that the killer is picked **before** it is executed.  
Otherwise, `mystery.killer` is `""`  because that is the default value.

An appropriate place to call this function might be after line 311, due to this.

code/clue.js
```diff
@@ -1,59 +1,396 @@
...
+  document.getElementById('weaponImage').src = `${mystery.weapon.image}`
+  weaponIsPicked = true
+    }, 2500)
+  }
+};
+document.getElementById('weaponCard').addEventListener('click', pickWeapon);
+
+//If pickKiller AND pickWeapon function is excecuted/cards are clicked then excecute this function on the pickroom:
+const pickRoom = () => {
+  if (killerIsPicked && weaponIsPicked) {
```
puj: Do you want to be able to pick the room again if it's already picked?
`  if (killerIsPicked && weaponIsPicked && !roomIsPicked) {`

code/clue.js
```diff
@@ -1,59 +1,396 @@
...
+
+// STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. 
+const revealMystery = (mystery) => {
+  //Invoke when button is clicked and show elements from mystery object in text:
+document.getElementById('mystery').innerHTML = `The murder was committed by ${mystery.killer.firstName} in the ${mystery.room.name} with ${mystery.weapon.name}` 
+document.getElementById('revealButton').innerHTML = `PLAY AGAIN?` //change the text on button.
 }
 
-// CREATE FUNCTIONS pickWeapon and pickRoom in a similar way.
+//BELOW I TRIED TO MAKE A FUNCTION SO THAT THE PAGE RELOADS WHEN 'PLAY AGAIN'-BUTTON IS CLICKED. IT DIDN'T WORK. I wanted to leave this here just to show the code-reviewer what I tried and maybe get pointers on how to achieve this.
```
puj: You had some good attempts here, but we can use your logic without adding a separate `onclick` handler

```javascript
const revealMystery = (mystery) => {
  // If the text is "PLAY AGAIN?" then the mystery has already been revealed
  if (document.getElementById('revealButton').innerText === 'PLAY AGAIN?') {
    location.reload();
  } else {
  // The text has not been changed yet, so we are revealing the mystery
    document.getElementById(
      'mystery'
    ).innerHTML = `The murder was committed by ${mystery.killer.firstName} in the ${mystery.room.name} with ${mystery.weapon.name}`;

    // Update the button text to reflect the action that can now be performed 
    document.getElementById('revealButton').innerText = 'PLAY AGAIN?'; //change the text on button.
  }
};
```

And, since we use the same handler function `revealMystery` we don't need any other functions

puj (approved):
This was a beautiful PR. 

@Hanna-Skeppe You had brilliant reflections and you highlighted the limitations of your code with the intention of enlightening the reviewer to your efforts.  This really helped the reviewer pinpoint where the feedback was most desired.  Your code comments were also very good at guiding the reviewer.  This is the core concept behind a PR.  

@the-j-curl You provided great feedback in the requested places and beyond that brought your own insight into the feedback.  

This is exactly what a PR should be, a discussion about how this code can be taken to the next level.  Great job to you both.

Hanna-Skeppe:
Thank you both for the review @puj @the-j-curl ! I will look in to the suggested changes and improvements 👍

code/clue.js
Hanna-Skeppe: Yes! Good point. Thank you! I noticed that the room-card was clickable also after first click although it shouldn't have been.

code/clue.js
Hanna-Skeppe: Thank you! Now it is working :)

code/clue.js
Hanna-Skeppe: Fixed the increaseChanceOfFavorite-function now so it works. I call this function in the pickWeapon-function.

code/clue.js
Hanna-Skeppe: I made it work now with help from Van and Max :)

code/clue.js
Hanna-Skeppe: Changed this now as you suggested 👍

code/style.css
Hanna-Skeppe: Fixed it now :)

code/style.css
Hanna-Skeppe: Yes, I could also have put the box-shadow higher up in CSS where I style:
.killer-deck,
.weapon-deck,
.room-deck

But as I have the same box-shadow on the button also I chose to target .cards article here, which is the same as: 
.killer-deck,
.weapon-deck,
.room-deck

In this way I didn't have to specify the box-shadow in two places. But maybe it isn't ideal to do as I chose?

code/index.html
Hanna-Skeppe: @the-j-curl I think I added the card container to be able to set position: relative to that element as it being the parent element to the loader element which has position absolute. It didn't work to set position relative on the article element since it is not a parent element to the loader.

code/index.html
Hanna-Skeppe: Removed the height-property from HTML as I had already specified the height of the images in CSS 👍

code/style.css
Hanna-Skeppe: Actually, I saw that I use both of the animations. One is for the animation on the cards, and the other is for the animation on the button.

PR #122, opened 2020-09-20
Cluedo mystery picker

LillanEriksson:
Aimed for red level this project.


README.md
```diff
@@ -1,13 +1,15 @@
...
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+This project is to make a Cluedo random mysterypicker. The goal is to practice using arrays, objects and loops.
 
 ## The problem
 
 Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
 
+I started with coding the javascript part, and making all the functions on the page work. After that I started with the styling of the page. This project I did not use Trello for planning as much as in the earlier projects, and to next assignment I will go back to my planning-board in Trello.
```
luuf81: Great idea to start ideation in Trello. Agree with re-usable functions. It´s something that takes a bit of time but valuable for readability of code for oneself and others :)

code/index.html
```diff
@@ -8,30 +8,47 @@
...
+        <div id="loader-wrapper" class="load-wrapper">
+          <div id="load2" class="loader">
+            <!--this is loader-->
+          </div>
+        </div>
       </article>
       <article class="room-deck" id="roomCard">
-        <p>ROOM</p>
+        <p id="room-headline" class="room-card-headline"></p>
         <p id="roomName"></p>
```
luuf81, line 40: Clear namings overall on classes and ids, only thing you could think about is to be consistent in using camel/kebab case for readbility

code/style.css
```diff
@@ -9,50 +11,141 @@ p {
   background-color: #fff;
 }
 
+.header-styling {
```
luuf81: I understand what it means but I would choose another naming convention on the class here, maybe just `.header`?

code/style.css
```diff
@@ -9,50 +11,141 @@ p {
...
+
+@keyframes spin {
+  0% {
+    transform: rotate(0deg) rotateY(0deg);
+  }
+
+  100% {
+    transform: rotate(360deg) rotateY(360deg);
+  }
+}
```
luuf81, line 153: really nice animation!

code/style.css
luuf81: Not a requirement, but the positioning of the header was a bit off in my browser. Maybe center the header so that it shows correctly when making wider/narrower

code/style.css
```diff
@@ -9,50 +11,141 @@ p {
...
+  background: rgba(10, 11, 13, 0.7);
   text-align: center;
   padding-top: 30px;
   border-radius: 16px;
+  box-shadow: 0 4px 8px 0 rgba(27, 107, 29, 0.2),
+    0 6px 20px 0 rgba(221, 217, 217, 0.4);
+  transition: transform 0.2s;
+}
+
+.room-deck:hover {
```
luuf81, line 87: This property could have been put together with killer-deck:hover and weapon-deck hover since they do the exact same thing.

code/style.css
```diff
@@ -9,50 +11,141 @@ p {
...
+  justify-content: center;
+}
+
+.room-deck .load-wrapper {
+  margin-top: 20px;
+}
+
+.weapon-deck .load-wrapper {
+  margin-top: 20px;
 }
```
luuf81, line 166: Overall I think your styling makes sense, it´s readable and you use the correct syntax / properties to adjust the styling according to the needs. I really like the design as well. Good work!

code/clue.js
```diff
@@ -3,57 +3,274 @@
...
+  killer: null,
+  weapon: null,
+  room: null,
+};
+//function to shuffle the suspects favourite weapon
+const randomFavouriteWeapon = () => randomSelector(weapons);
+
+const killerLoader = document.getElementById('load');
+const weaponLoader = document.getElementById('load2');
+const roomLoader = document.getElementById('load3');
```
luuf81, line 178: Good that you are assigning these elements to variables, easier on the eyes!

code/clue.js
```diff
@@ -3,57 +3,274 @@
...
+const killerHeadLine = document.getElementById('killer-headline');
+const weaponHeadLine = document.getElementById('weapon-headline');
+const roomHeadLine = document.getElementById('room-headline');
+
+//functions to show loaders and change the before the information
+const showKillerLoader = () => {
+  killerHeadLine.innerHTML = 'Loading killer-information..';
+  killerLoader.style.display = 'block';
+};
+const showWeaponLoader = () => {
```
luuf81, line 189: Good naming overall on Functions => I understand what they do :)

code/clue.js
```diff
@@ -3,57 +3,274 @@
...
-  ).innerHTML = `${mystery.killer.firstName} ${mystery.killer.lastName}`
-}
+  showKillerLoader();
+  setTimeout(killerInformation, 4000);
+};
 
 // CREATE FUNCTIONS pickWeapon and pickRoom in a similar way.
+const pickWeapon = () => {
+  showWeaponLoader();
+  setTimeout(weaponInformation, 4000);
```
luuf81, line 252: setTimeout works well in this case because we have set a time for how long the animation is running. For future projects you could try out callback functions, meaning sending a function as a parameter to another function, which then ensures that function nr2 will only execute once the first one is finished...

code/clue.js
```diff
@@ -3,57 +3,274 @@
...
+};
+
+document.getElementById('killerCard').onclick = pickKiller;
+document.getElementById('weaponCard').onclick = pickWeapon;
+document.getElementById('roomCard').onclick = pickRoom;
+
+//The first that shows on the cards
+killerHeadLine.innerHTML = 'Pick a killer';
+weaponHeadLine.innerHTML = 'Pick a weapon';
+roomHeadLine.innerHTML = 'Pick a room';
```
luuf81, line 267: For logical order It would have made more sense if these lines came in the beginning of the js-file.

code/clue.js
```diff
@@ -3,57 +3,274 @@
...
 
 // STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
 // 'The murder was committed by Jacob Green, in the living room with a rope.'
+const revealMystery = () => {
+  document.getElementById(
+    'mystery'
+  ).innerHTML = `The murder was commited by ${mystery.killer.firstName} ${mystery.killer.lastName} in the ${mystery.room} with a ${mystery.weapon.name}.`;
+};
+document.getElementById('revealButton').onclick = revealMystery;
```
luuf81, line 286: Not a requirement but could have been nice with some validation when I click the button :)

luuf81, line 286: Overall I think your js-code is very well put together. It´s easy to read, I understand the intentions of the code, and it does what´s it supposed to do. Great work!

luuf81 (approved):
Great job Lillemor! It was a joy reading your code since it was easy to comprehend and structured. You aimed for red level which I think you have achieved!

I have made some comments for things to think about, but I have no change requests :)

README.md
```diff
@@ -1,13 +1,15 @@
...
+
+If I had more time I would have written some of the functions to be re-usable.
```
puj: 👍

code/clue.js
puj: Very much agreed!

code/clue.js
```diff
@@ -3,57 +3,274 @@
...
+const weaponHeadLine = document.getElementById('weapon-headline');
+const roomHeadLine = document.getElementById('room-headline');
+
+//functions to show loaders and change the before the information
+const showKillerLoader = () => {
+  killerHeadLine.innerHTML = 'Loading killer-information..';
+  killerLoader.style.display = 'block';
+};
+const showWeaponLoader = () => {
+  weaponHeadLine.innerHTML = 'Loading weapon-information..';
```
puj, line 190: Super nice touch :)

code/clue.js
```diff
@@ -3,57 +3,274 @@
...
+  );
+
+  mystery.killer = randomSelector(suspects);
+  killerLoader.style.display = 'none';
+  killerHeadLine.innerHTML = 'Killer-information:';
+  killerCard.style.background = mystery.killer.color;
+  killerName.innerHTML = `${mystery.killer.firstName} ${mystery.killer.lastName}`;
+  killerImage.src = mystery.killer.image;
+  killerAge.innerHTML = `${mystery.killer.age} years old`;
+  killerOccupation.innerHTML = `${mystery.killer.occupation}`;
```
puj, line 216: This could be simply
` killerOccupation.innerHTML = mystery.killer.occupation;`

code/clue.js
puj: This is also worth checking out in this case https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onanimationend

code/clue.js
puj: I can agree with this 👍

code/clue.js
puj: Agreed, we should not have errors.
This happens when we don't click every card, but we try to reveal anyway.  Let's fix that @LillanEriksson 
image

puj: Agreed

code/style.css
puj: `styling` could be omitted here, indeed

code/style.css
puj: It also seems like the `.room-deck` and `.weapon-deck` and `.killer-deck` have similar CSS properties

code/style.css
puj: One of my favorites!

code/style.css
puj: Well said!

puj (changes requested):
Really nice work @LillanEriksson 

I can only echo what @luuf81 pointed out in several places, and overall very well structured code.  💯   


Changes Requested:
 - Add some validation to the `revealMystery` function to prevent the error from occuring as noted in the comments

code/clue.js
LillanEriksson: Fixed it now!

puj (approved):
Looks good 👍 Approved!

PR #124, opened 2020-09-20
Cluedo Project - Marjaana Tienaho

MayaBlixt:


clue.js
```diff
@@ -0,0 +1,193 @@
...
+  occupation: 'Entrepreneur'
+};
+const proessorPlum = {
+  firstName: 'Victor',
+  lastName: 'Plum',
+  color: 'purple',
+  description: 'He is good at lying',
+  age: 29,
+  image: 'assets/plum.png',
+  occupation: 'Developer'
```
Corneliaryas, line 18: Nice information of the suspects, especially descriptions and occupations!

clue.js
```diff
@@ -0,0 +1,193 @@
...
+  'Spa',
+  'Living room',
+  'Observatory',
+  'Theater',
+  'Guest house',
+  'Patio'
+]
+console.log(mrGreen.description)
+console.log(rope.name)
+console.log(rooms.toString())
```
Corneliaryas: Great to check the console log while coding, maybe a bit cleaner code if removed when deployed?

clue.js
```diff
@@ -0,0 +1,193 @@
...
+  mystery.room = randomSelector(rooms)
+
+  document.getElementById('roomName').innerHTML = `Room where the crime happend: ${mystery.room}`
+}
+document.getElementById('roomCard').addEventListener('click', pickRoom)
+
+//Function that reveals the mystery
+const revealMystery = () => {
+  console.log(`The murder was commited by ${mystery.killer.firstName} ${mystery.killer.lastName}in the ${mystery.room} with a ${mystery.weapon.name} `)
+  document.getElementById('mystery').innerHTML = `The murder was commited by ${mystery.killer.firstName} ${mystery.killer.lastName}in the ${mystery.room} with a ${mystery.weapon.name} `
```
Corneliaryas, line 181: Super small thing: add a space after ${mystery.killer.lastname} so the words get separated ;)

index.html
```diff
@@ -0,0 +1,54 @@
...
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+  <!--Mystery favicon-->
+  <link href="./assets/favicon.ico" rel="icon" type="image/x-icon" />
+  <!--Google fonts-->
+  <link href="https://fonts.googleapis.com/css?family=Arvo|Lato" rel="stylesheet">
+  <link rel="stylesheet" href="style.css" />
+  <title>Mystery cluedo project</title>
```
Corneliaryas, line 13: Nice with favicon and google fonts!

index.html
```diff
@@ -0,0 +1,54 @@
...
+  <h2>Choose a killer, a weapon and a room and then reveal the crime </h2>
+  <section class="cards">
+    <article class="killer-deck" id="killerCard">
+      <h2>KILLER</h2>
+      <p id="killerName"></p>
+      <p id="killerAge"></p>
+      <p id="killerOccupation"></p>
+      <p id="killerDescription"></p>
+      <div class="killerImg-container" id="killerImgContainer">
+        <img id="killerImage" src="./assets/icons8-inquiry-100.png" height="100" src="" alt="Killer" />
```
Corneliaryas: Super nice with custom icons! 👍

index.html
```diff
@@ -0,0 +1,54 @@
...
+    </article>
+  </section>
+
+  <h2 id="mystery"></h2>
+  <div class="buttonDiv">
+    <button id="revealBtn">Reveal the Crime</button>
+  </div>
+  <footer class="footer">
+    <p> © 2020 Marjaana Tienaho. Project Week 5 (Project Cluedo) during the Technigo Frontend Bootcamp Autumn 2020.</p>
+  </footer>
```
Corneliaryas, line 50: Great do have this extra info with a footer!

style.css
```diff
@@ -0,0 +1,118 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  font-family: "Courier New", Courier, monospace;
+  background-image: url("./assets/house.jpg");
```
Corneliaryas, line 8: Great background image! On larger screens, it repeats now. If you want to avoid that I think you could add something like: `background-size: cover;
  background-repeat: no-repeat;`

style.css
```diff
@@ -0,0 +1,118 @@
...
+  height: 400px;
+  text-align: center;
+  padding-top: 30px;
+  border-radius: 16px;
+}
+
+.buttonDiv {
+  display: block;
+  width: 400px;
+  margin: 0 auto 0 auto;
```
Corneliaryas: If you want to make this shorter I think `margin: 0 auto;` does the same thing :)

style.css
```diff
@@ -0,0 +1,118 @@
...
+
+@keyframes background {
+  0% {
+    background-position: 0 0;
+  }
+
+  100% {
+    background-position: 400px 0;
+  }
+}
```
Corneliaryas, line 115: Wow this hover animation is so great! Loved the effect when I saw it on the site! 😍

style.css
```diff
@@ -0,0 +1,118 @@
...
+  align-items: center;
+}
+
+img {
+  padding-top: 30px;
+}
+
+.killer-deck,
+.weapon-deck,
+.room-deck {
```
Corneliaryas, line 52: Smart move to line them up like this instead of repeating.  I can't see that they are altered or used in a different way later on though, could they be the same class altogether? (like .card-deck or something) 
In JS you call the killer decks ID and alter the style, but I think that works even if the class is the same. Just food for thought :)

style.css
```diff
@@ -0,0 +1,118 @@
...
+
+  100% {
+    background-position: 400px 0;
+  }
+}
+
+footer {
+  color: white;
+  padding: 20px;
+}
```
Corneliaryas, line 120: Great flow in the CSS file, I think the order and grouping of classes make perfect sense and make it easy to read!

clue.js
```diff
@@ -0,0 +1,193 @@
...
+
+}
+document.getElementById('revealBtn').addEventListener('click', revealMystery)
```
Corneliaryas, line 188: Great order of declarations with objects and arrays first and then functions. Maybe group the invoking (the click events) of the functions at the end too? Just to make it even easier to follow along :)

index.html
```diff
@@ -0,0 +1,54 @@
...
+      <img id="weaponImage" src="./assets/icons8-weapons-100.png" height="100" src="" alt="Weapon" />
+    </article>
+    <article class="room-deck" id="roomCard">
+      <h2>ROOM</h2>
+      <p id="roomName"></p>
+      <img id="weaponImage" src="./assets/icons8-prefab-house-100.png" height="100" src="" alt="House" />
+    </article>
+  </section>
+
+  <h2 id="mystery"></h2>
```
Corneliaryas, line 44: Love the style, just a little hard to read the mystery text here, maybe add a background or change the font color. I always struggle with this when having text over image ;)

Corneliaryas (approved):
Great job Marjaana! 😃 You write clear and easy to follow code I think, both in HTML, CSS and JS. Super nice touch with the icons and the hover animation on the button. I learned a lot by looking at your code that I will try to implement on future projects. Good luck with the next one! :D

clue.js
MayaBlixt: Thanks :)

clue.js
MayaBlixt: Fixed, you are right! I thought I removed them all but apparently not 🤣

clue.js
MayaBlixt: Fixed :)

index.html
MayaBlixt: Thanks :)

style.css
MayaBlixt: Thanks! Tried your suggestions and it worked perfectly :)

style.css
MayaBlixt: Fixed as suggested :)

style.css
MayaBlixt: :)

style.css
MayaBlixt: Great suggestion :)

index.html
MayaBlixt: I agree :) Changed the font color to red.

clue.js
MayaBlixt: Moved the invoking functions to the end :)

clue.js
```diff
@@ -0,0 +1,189 @@
...
+// Array of suspects
+const suspects = [
+  mrGreen,
+  proessorPlum,
+  missScarlet,
+  mrsPeacock,
+  colonelMustard,
+  mrsWhite
+]
+suspects.forEach((element) => console.log(element));
```
EmOlsz: Delete console.logs before final push

clue.js
```diff
@@ -0,0 +1,189 @@
...
+// Function that randomly selects one item from the array
+const randomSelector = array => {
+  return array[Math.floor(Math.random() * array.length)]
+}
+
+// Object with a killer, a weapon and a room  
+let mystery = {
+  killer: 'unknown',
+  weapon: 'unknown',
+  room: 'none'
```
EmOlsz: It's far better approach to use `null` values for initialisation :
```
const mystery = {
  killer: null,
  weapon: null,
  room: null
}
```
Moreover, you can use `const` - we will be updating value of this variable, that's true, but since it will still be object, JavaScript will not consider property changes as overwriting value and therefore won't throw error

clue.js
```diff
@@ -0,0 +1,189 @@
...
+// Function that picks a room
+const pickRoom = () => {
+  mystery.room = randomSelector(rooms)
+
+  document.getElementById('roomName').innerHTML = `Room where the crime happend: ${mystery.room}`
+}
+
+//Function that reveals the mystery
+const revealMystery = () => {
+  console.log(`The murder was commited by ${mystery.killer.firstName} ${mystery.killer.lastName} in the ${mystery.room} with a ${mystery.weapon.name} `)
```
EmOlsz: Delete all console logs from your code

code/clue.js
```diff
@@ -8,28 +8,142 @@ const mrGreen = {
   age: 45,
   image: 'assets/green.png',
   occupation: 'Entrepreneur'
-}
+};
+
+const proessorPlum = {
```
EmOlsz, line 13: You don't need whole this file, right? Let's delete it

index.html
```diff
@@ -0,0 +1,54 @@
...
+      <div class="killerImg-container" id="killerImgContainer">
+        <img id="killerImage" src="./assets/icons8-inquiry-100.png" height="100" src="" alt="Killer" />
+      </div>
+
+    </article>
+    <article class="weapon-deck" id="weaponCard">
+      <h2>WEAPON</h2>
+      <p id="weaponName"></p>
+      <p id="weaponWeight"></p>
+      <img id="weaponImage" src="./assets/icons8-weapons-100.png" height="100" src="" alt="Weapon" />
```
EmOlsz: You duplicate `src` attribute

style.css
```diff
@@ -0,0 +1,120 @@
...
+  background-image: url("./assets/house.jpg");
+  background-size: cover; 
+  background-repeat: no-repeat;
+}
+
+p {
+  margin: 1px;
+  padding-top: 5px;
+  font-weight: 300;
+  Font-Family: 'Lato', Sans-Serif;
```
EmOlsz: it's `sans-serif`, not `Sans-Serif` :)

style.css
EmOlsz: Very good suggestion!

style.css
```diff
@@ -0,0 +1,120 @@
...
+      transparent 10%,
+      rgb(70, 67, 67) 10% 20%,
+      transparent 20% 30%,
+      rgb(70, 67, 67) 30% 40%,
+      transparent 40% 50%,
+      rgb(70, 67, 67) 50% 60%,
+      transparent 60% 70%,
+      rgb(70, 67, 67) 70% 80%,
+      transparent 80% 90%,
+      rgb(70, 67, 67) 90% 100%);
```
EmOlsz, line 103: Nice line break :)

EmOlsz (changes requested):
Good job Marjaana! Your code is pretty solid and consequent :) There are some things I would like you to fix, please take a look on my comments and provide changes. After that, I will be able to approve your project

clue.js
MayaBlixt: Fixed :)

clue.js
MayaBlixt: Good point! Updated accordingly above :)

clue.js
MayaBlixt: Removed

code/clue.js
MayaBlixt: Sorry, but do not understand the above comment.🤔   Could you elaborate ?

index.html
MayaBlixt: You´r right! Removed extra scr attribute

style.css
MayaBlixt: Ah, changed to sans-serif.

MayaBlixt (commented):
Made updates as suggested above

code/clue.js
EmOlsz: I meant that you have 2 `clue.js` files in your project, but this one has only default code provided by Technigo as a starter

EmOlsz (approved):

code/clue.js
MayaBlixt: Ah now I understand, I removed the extra js file 😄

PR #127, opened 2020-09-20
Cluedo powered by FRIENDS 

lilihorvath:


README.md
```diff
@@ -1,13 +1,11 @@
...
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+So, our job was that build a mystery picker to find something or someone and pick randomly. 
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+The first problem to find out how to make functions what are working together and actually it was quite helpful that I grabbed paper and pen and draw/write how I imagine the problem and fix it. So sometime go back to the past and dont be afraid of using paper and pen.(:
```
destinyleia, line 7: Love that you used paper and pen! That's a makes a really god overview - great idea! 👌🏻

code/clue.js
```diff
@@ -1,59 +1,237 @@
...
+
+const RossApartment = {
+  name: "Ross' Apartment",
+  image: 'assets/ra.jpg'
+}
 
-const rooms = []
+const ChandlerDiningRoom = {
+  name: "Chandler's Dining Room",
+  image: 'assets/cdr.jpg'
```
destinyleia: Not a huge problem but remember that Maks told us to pick either " or ' to write in our code to make it more clean using one of them. 😁

code/clue.js
```diff
@@ -1,59 +1,237 @@
...
+document.getElementById('friendsCard').onclick = pickFriends
 
-  // This will change the background color of the card to the one connected to the chosen killer and show the full name of the killer. Feel free to add more things to show about the killer.
-  document.getElementById('killerCard').style.background = mystery.killer.color
-  document.getElementById(
-    'killerName'
-  ).innerHTML = `${mystery.killer.firstName} ${mystery.killer.lastName}`
+const pickFoods = () => {
+  mystery.foods = randomSelector(foods)
+  document.getElementById('foodsName').innerHTML = mystery.foods.name;
```
destinyleia: For having a more clean code u should stick with writing the code with the ${ } as in 224 in every code that writing like that. For a better structure ❤️

code/clue.js
```diff
@@ -1,59 +1,237 @@
...
+  document.getElementById('roomsImage').src = mystery.rooms.image
+}
 
-// STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
-// 'The murder was committed by Jacob Green, in the living room with a rope.'
+document.getElementById('roomsCard').onclick = pickRooms
+
+const revealMystery = () => {
+  if (mystery.friends === undefined || mystery.foods === undefined || mystery.rooms === undefined ){
+    document.getElementById('mystery').innerHTML = "You must find Joey's food!"
```
destinyleia, line 232: I like that you put in "You must find Joey's  food!" when I haven't picked any card. That was creative! 😄

code/style.css
```diff
@@ -1,58 +1,140 @@
...
+  background-image: url(./assets/background.jpg);
+  background-size: cover;
 }
 
-p {
-  margin: 1px;
-  font-weight: 300;
-  background-color: #fff;
+.main-container {
+  min-height: 100vh;
```
destinyleia, line 23: what's vh? 😁

code/style.css
```diff
@@ -1,58 +1,140 @@
...
+
+footer p {
+  text-align: center;
+  margin: 30px 0 0 0;
+}
+
+a {
+  text-decoration: none;
+  color: #000000;
+}
```
destinyleia, line 136: I LOVE THAT YOU WENT YOUR OWN WAY WITH A FRIENDS THEME! 😍 A FOR EFFORT! 💯😇❤️

code/style.css
```diff
@@ -1,58 +1,140 @@
...
+  margin: 2px;
+  font-weight: 300;
+  font-size: 15px;
 }
 
-.room-deck {
-  width: 250px;
-  height: 400px;
-  background: midnightblue;
+#friendsImage, #foodsImage, #roomsImage {
```
destinyleia: I don't know if this is necessary with the # but not sure, correct me if Im wrong please! 😁

destinyleia (approved):
Wow Lili! Love that you did it with an inspiration from Friends! 😄 It was really funny and creative of you! I comment of some things that I thought could help you - like a tip! But I think your code looked very nice and structured and that it was easy to read! Great job Lili and keep it up by doing it YOUR WAY! 😍

code/clue.js
lilihorvath: Thank you Destiny, I am going to fix it!(:

code/style.css
lilihorvath: https://stackoverflow.com/questions/41793228/min-height-in-vh-vs-for-body/41793897

code/style.css
lilihorvath: I have used because I did not want to do one more different class for the images!(:

code/clue.js
lilihorvath: Thank you Destiny!

code/clue.js
lilihorvath: Yes, you're totally right, I am going to fix it! Thanks a lot!!!

code/clue.js
```diff
@@ -1,59 +1,237 @@
-// STEP 1 - CREATE OBJECTS FOR ALL THE SUSPECTS, SOMETHING LIKE THIS:
+const Marcel = {
+  firstName: 'Marcel',
+  lastName: 'the Monkey',
```
EmOlsz: Congratulations for making this project original :)
Just one thing - we shouldn't declare variables with capital letters. This approach is reserved in JavaScript for classes. Needs to be fixed

code/clue.js
```diff
@@ -1,59 +1,237 @@
...
+  age: 27,
+  image: 'assets/ross.png',
+  occupation: 'Palaeontologist',
+  favouriteFood: 'Muffin'
+};
+
+const Chandler = {
+  firstName: 'Chandler',
+  lastName: 'Muriel Bing',
+  description: 'Making jokes whe he feels himself uncomfortable and loves Monica.',
```
EmOlsz: Spoiler alert!

code/clue.js
EmOlsz: Exactly :)

code/clue.js
```diff
@@ -1,59 +1,237 @@
...
   return array[Math.floor(Math.random() * array.length)]
 }
 
-// CREATE AN OBJECT THAT KEEPS THE MYSTERY.
-// With a killer, a weapon and a room.
-// The values will be set later.
+const mystery = {
+  friend: undefined,
+  foods: undefined,
+  rooms: undefined
```
EmOlsz: It's even more semantic approach to initialise mystery object with null values like this :

```
const mystery = {
  friend: null,
  foods: null,
  rooms: null
}
```

code/clue.js
EmOlsz: Not really - we should use template strings syntax (`${}`) only when we want to combine variable with fixed string, like in 
`document.getElementById('foodsName').innerHTML = 'Mystery food was ${mystery.foods.name;}'`

On the other hand, if you only need to pass variable without any fixed text like `"Mystery food was"`, then you can (and you even should) just write :
`document.getElementById('foodsName').innerHTML = mystery.foods.name;`

code/clue.js
```diff
@@ -1,59 +1,237 @@
...
+  document.getElementById('roomsName').innerHTML = `${mystery.rooms.name}`
+  document.getElementById('roomsImage').src = mystery.rooms.image
+}
 
-// STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
-// 'The murder was committed by Jacob Green, in the living room with a rope.'
+document.getElementById('roomsCard').onclick = pickRooms
+
+const revealMystery = () => {
+  if (mystery.friends === undefined || mystery.foods === undefined || mystery.rooms === undefined ){
```
EmOlsz: We can write this line shorter like that :

`if (!mystery.friends || !mystery.foods || !mystery.rooms ) { ... }`

For further information check https://www.freecodecamp.org/news/js-type-coercion-explained-27ba3d9a2839/

code/style.css
```diff
@@ -1,58 +1,140 @@
...
+body {
+ height: 100%;
+ position: relative;
+}
+
 body {
   margin: 0;
-  font-family: "Courier New", Courier, monospace;
+  font-family: GabrielWeissFriendsFont;
+  background-image: url(./assets/background.jpg);
```
EmOlsz: It's far more safe approach to wrap path in parentheses. Without them, some browsers might not find the path
`background-image: url("./assets/background.jpg");`

code/style.css
```diff
@@ -1,58 +1,140 @@
...
+  margin-bottom: 0;
+  font-family: GabrielWeissFriendsFont;
+  color: #000000;
+  text-align: center;
+}
+
+h2 {
+  margin-top: 0;
+  font-family: GabrielWeissFriendsFont;
+  color: #000000;
```
EmOlsz: We can write it shorter: `#000`;

code/style.css
EmOlsz: Honestly, you should Lili :) In CSS we should use only classes, no id's

code/style.css
```diff
@@ -1,58 +1,140 @@
...
   border-radius: 16px;
+  position: relative;
 }
 
 button {
-  padding: 8px 16px;
-  background: crimson;
+  padding: 10px 18px;
+  background: black;
+  color: #ffffff;
```
EmOlsz: `#fff`

code/style.css
```diff
@@ -1,58 +1,140 @@
...
-  margin: 0 auto;
+  margin: 0 auto 30px auto;
+  font-family: GabrielWeissFriendsFont;
 }
+
+#restartButton {
+  display: none;
+}
+
+article, #cookButton, #refresgPage {
```
EmOlsz: change id's for classes

EmOlsz (changes requested):
Congrats Lili :) Good job this week, especially for your own interpretation of project :)
Nonetheless, there are some things in your code which needs to be fixed. Please, take a look on my comments and add changes. Only after that I'll be able to approve your project

code/clue.js
lilihorvath: I am on it!

code/clue.js
lilihorvath: If you have ever seen just one minute of friends it soo obvious that they are in love, even the scene is not about Monica and Chandler:D

code/clue.js
lilihorvath: Hmm, I understand, but is it truly necessary? I mean, yes, I could use null definitely, but why undefined is not perfect here? Sorry for the stupid question.

code/clue.js
lilihorvath: Ooo, wow! Yes, and it is more cleaner! Than you and that was a great article of it!

code/style.css
lilihorvath: Yess, I know, and I am going to fix that, but I love memorizing colors this way.

code/style.css
lilihorvath: I am on it!

code/style.css
lilihorvath: Yess, I understand I have just seen in a video on youtube and that looked soo fancy.:/

code/style.css
lilihorvath: Definitely, I totally forget that!

lilihorvath:
Fixed. Honestly, I loved the way how you left comments and told me what changes needs in my project. I understands the change's whys, and I have to say thank you for the hints too! Thank you @EmOlsz !

code/clue.js
EmOlsz: It's a convention - developers have a lot of them and this is one of them. Thanks to that we can communicate without words. If I see `mystery` object with `null` values, I'm 100% sure that somewhere in your code later on you will update those values and for now you only want to indicate that initially they are empty.
`undefined` on the other hand can only lead to confusions and wrong assumptions done by other developers. It's because if you see `undefined` in your code, in 99% of cases it's because function returned something in wrong way (like element from array at wrong index which doesn't exists), or didn't return anything (so it returned `undefined`).
As you see, at that this point we are no longer sure who've done `undefined` value - was it developer, or maybe program itself during execution? A lot of questions which we can avoid by sticking to convention of `null`. JavaScript will never return `null` by itself

EmOlsz:
> Fixed. Honestly, I loved the way how you left comments and told me what changes needs in my project. I understands the change's whys, and I have to say thank you for the hints too! Thank you @EmOlsz !

It's a pleasure for me Lili! You've implemented most of stuff, good job :) Before acceptance, I'll still wait for lowercase names of variables from top of the file. Moreover, I see you've changed `undefined` to `null`, but in the same moment you changed your properties from lowercase to uppercase (like `friends` to `Friends`). They need to be lowercase as well, so change that

lilihorvath:
I hope that I understood well, I am a bit confused now...and sorry for the late answer!

EmOlsz (approved):

PR #137, opened 2020-09-20
Project 5: Cluedo Mystery Picker

Thessan:


code/clue.js
```diff
@@ -1,59 +1,202 @@
...
-// CREATE AN OBJECT THAT KEEPS THE MYSTERY.
-// With a killer, a weapon and a room.
-// The values will be set later.
+// AN OBJECT THAT KEEPS THE MYSTERY.
 
-// This function will be invoked when you click on the killer card.
+const mystery = {
+  killer: null,
+  weapon: null,
+  room: null
```
LillanEriksson, line 158: Great that you set the value to null to keep from exposing the mystery before all cards are clicked.

code/index.html
```diff
@@ -5,23 +5,29 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <link rel="stylesheet" href="style.css" />
-    <title>Technigo Project Cluedo</title>
+    <link rel="icon" type="img/png" href="./Images/spy.png">
```
LillanEriksson, line 8: Nice favicon!

code/index.html
```diff
@@ -30,8 +36,13 @@
 
     <h1 id="mystery"></h1>
 
-    <button>REVEAL THE CRIME</button>
+    <button onclick="revealMystery()" id="clickMe">Reveal the crime</button>
```
LillanEriksson: I think Maks or Van (really unsure who of them) once said that you should put the id-attribute first, then class, then the other attribute-stuff.

code/index.html
```diff
@@ -30,8 +36,13 @@
...
+
+    <footer>
+      <img id="logo" src="./Images/logo.svg">
+        <p id="footer">Created by Therése Larsson / Project Week 5</p>
+    </footer>
```
LillanEriksson, line 44: Very nice footer!

code/style.css
```diff
@@ -1,58 +1,126 @@
...
+  background-size: cover;
+}
+
+
+h1 {
+  margin-top: 100px;
+  font-family: 'Grenze Gotisch', cursive;
+  font-size: 60px;
+  color: rgb(255, 255, 255);
+  text-shadow: 5px 5px 5px rgba(128, 128, 128, 0.753);
```
LillanEriksson, line 15: Love the text-shadow, very subtle.

code/style.css
```diff
@@ -1,58 +1,126 @@
...
+  text-align: center;
 }
 
-p {
-  margin: 1px;
+
+#mystery {
```
LillanEriksson: Here you have mixed styling classes and id's, maybe it would be better to just keep to styling one or the other, you can always add a class to a html-element for the styling purpose, and let the id's be exclusive for the javascript.

code/clue.js
```diff
@@ -1,59 +1,202 @@
...
-  ).innerHTML = `${mystery.killer.firstName} ${mystery.killer.lastName}`
+// This will randomly select a killer from the suspects. And add that to the mystery object.
+mystery.killer = randomSelector(suspects);
+
+  document.getElementById('killerCard').style.background = mystery.killer.color;
+  document.getElementById('killerName').innerHTML = `Name: ${mystery.killer.firstName} ${mystery.killer.lastName}`;
+  document.getElementById('killerAge').innerHTML = `Age: ${mystery.killer.age}`;
+  document.getElementById('killerOccupation').innerHTML = `Occupation: ${mystery.killer.occupation}`;
+  document.getElementById('killerImage').src = mystery.killer.image;
 }
```
LillanEriksson, line 172: I did put all these in variables, I think it made the code look very neat and became so much more readable. Like this:
`const killerCard = document.getElementById('killerCard')`
and then when I want to change the killerCard element 
`killerCard.style.background = mystery.killer.color`

LillanEriksson (approved):
The page looks good, and my opinion is that you passed the goal for the blue level. I believe you probably could aim higher then blue, you are better than you think you are!

To the next project you could think of being consistent when writing the html-attributes, ex firdt id, then class and so on, like I mentioned in one of my comments.

code/style.css
Thessan: thanks, good point! =)

code/clue.js
Thessan: thanks, I'll have a look at your code and compare a bit, see if I understand the differences =)

code/clue.js
```diff
@@ -1,59 +1,201 @@
...
-    'killerName'
-  ).innerHTML = `${mystery.killer.firstName} ${mystery.killer.lastName}`
+// This will randomly select a killer from the suspects. And add that to the mystery object.
+mystery.killer = randomSelector(suspects);
+
+  document.getElementById('killerCard').style.background = mystery.killer.color;
+  document.getElementById('killerName').innerHTML = `Name: ${mystery.killer.firstName} ${mystery.killer.lastName}`;
+  document.getElementById('killerAge').innerHTML = `Age: ${mystery.killer.age}`;
+  document.getElementById('killerOccupation').innerHTML = `Occupation: ${mystery.killer.occupation}`;
+  document.getElementById('killerImage').src = mystery.killer.image;
```
puj, line 171: Good job only using **`** where needed

code/clue.js
```diff
@@ -1,59 +1,201 @@
...
 }
 
-// CREATE FUNCTIONS pickWeapon and pickRoom in a similar way.
 
-// STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
-// 'The murder was committed by Jacob Green, in the living room with a rope.'
+// This function will be invoked when you click on the WEAPON CARD.
+  const pickWeapon = () => {
+
+  mystery.weapon = randomSelector(weapons);
```
puj, line 177: Indentation

code/clue.js
```diff
@@ -1,59 +1,201 @@
...
+
+  document.getElementById('killerCard').onclick = pickKiller;
+  document.getElementById('weaponCard').onclick = pickWeapon;
+  document.getElementById('roomCard').onclick = pickRoom;
+
+// FUNCTION revealMystery that will be invoked when you click that button.
+const revealMystery = () => {
+  document.getElementById('mystery').innerHTML = `The murder was committed by ${mystery.killer.firstName} ${mystery.killer.lastName}, with a ${mystery.weapon.name} in the ${mystery.room}` 
+  document.getElementById("clickMe").onclick = revealMystery;
+}
```
puj: In this function, we need some sort of validation or check to make sure all the `mystery` information is available before we attempt to use it
image

puj (changes requested):
Nice job @Thessan 
You are doing great! Nice job with the styles! I can echo what @LillanEriksson said 

> I believe you probably could aim higher then blue, you are better than you think you are!


Changes requested:
- Some indentation in the Javascript
- Validation in the `revealMystery` function

code/clue.js
Thessan: I know the reveal-button does not work if you don't choose all three cards, but not sure I understand what to fix here. Can you please clarify?

code/clue.js
Thessan: I'm a bit lost what this mean as well, is it that the info on row 178 should be placed on 176?

code/clue.js
Thessan: @puj Still need a bit of clarification on this 🙏

code/clue.js
Thessan: @puj and here, I'm unsure what you mean with Indentation

puj (approved):

PR #138, opened 2020-09-20
Cluedo mystery picker - Anna H

AnnaHellqvist:
**Link to deployed project:**
https://create-cluedo-mystery.netlify.app/

**Our reflections on how this week's project turned out:**
The assignment was good to create an understanding for objects and arrays and the difference between them and when to use which one. I am kind of happy with how my project turned out. I also felt like this project deepened my knowledge of functions. It was fun playing around with a loader and trying to get it to work the way I wanted to. I feel like one of my biggest struggles is the styling of the page, and this often takes a lot of time. 

**If I had more time I would have:**
- put some more effort into styling.
- added an animation to flip the cards when they are picked. I tried but felt I needed some more time to get it to work.
- made it responsive.
-added the function to make the favoriteweapon less random.. I probably will do this afterwards when I have some time.


code/style.css
```diff
@@ -1,58 +1,176 @@
...
+  font-size: 12px;
+  color: white;
+  background: black;
+  padding: 0 50px;
+}
+
+footer a {
+  text-decoration: none;
+  color: grey;
 }
```
stjernberg, line 175: //I love your styling.😍 It really fits in with the theme of the project. Great use of positioning with flexbox. Next time, rememberg to syle only by class-name. Use id only for javascrict. Also consider giving the selctors a classname for styling. I had feedback in my last project that styling by selectors is "bad practice."

code/style.css
```diff
@@ -1,58 +1,176 @@
...
+.revelation {
+  height: 100px;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: space-around;
+  margin-bottom: 10px;
+}
+
+#mystery {
```
stjernberg: Should be a classname.

code/clue.js
```diff
@@ -1,59 +1,319 @@
...
+  } else {
+    document.getElementById('mystery').innerHTML = 
+    `The murder was commited by ${mystery.killer.firstName} ${mystery.killer.lastName}, in the ${mystery.room} with a ${mystery.weapon.name}!`;
+  }
+}
 
-// STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
-// 'The murder was committed by Jacob Green, in the living room with a rope.'
+// Calling the revealMystery function
+document.getElementById('revealButton').onclick = revealMystery;
```
stjernberg, line 319: //Your code is well structured and easy to follow.  I like all the comment which helps the reader and yourself understand what the code is doing. Well done with adding extra stuff like shuffling random weapons and loading cards with set time. 👏You could shorten the code a little by combining arrays and objects. like const suspects = [{mrGreen-object},{prPlum.object}...]. Perhaps won't work with the weapons though, since thay are picked randomly from the array.

code/index.html
```diff
@@ -4,34 +4,67 @@
...
-    <button>REVEAL THE CRIME</button>
+    <footer>
+      <p>Project 4, Cluedo Mystery Picker, by Anna Hellqvist @ Technigo Bootcamp 2020</p>
+      <p><a target="_blank" href="https://icons8.com/icons/set/magnifying-glass-tilted-left-emoji">Magnifying Glass Tilted Left icon</a> icon by <a target="_blank" href="https://icons8.com">Icons8</a></p>
+    </footer>
 
+  
     <script src="clue.js"></script>
   </body>
 </html>
```
stjernberg, line 70: Nicely structured and clean code!✌ Just think about adding ./ first in the file path, like "./clue.js

stjernberg (commented):
Fantastic work with this week's project Anna 👏 Your site looks awesome and works well! I love the cool loading-animation that you added. That adds "the little extra" to the game. You really took on the challenges this week with extra randomising function and with your styling. Imagine how much you achieved already in these few weeks! Be proud of yourself 😍

stjernberg (approved):

code/clue.js
```diff
@@ -1,59 +1,319 @@
...
+
+const roomGuest = 'Guest House';
+
+const roomPatio = 'Patio';
+
+// OBJECT THAT KEEPS THE MYSTERY.
+
+let mystery = {
+  killer: '',
+  weapon: '',
```
EmOlsz: It's even more semantic approach to initialise mystery object with `null` values like this :

```
const mystery = {
  killer: null,
  room: null,
  weapon: null
}
```

Moreover, you can use `const` keyword - we will be changing values of properties, that's true, but JavaScript will not consider this as overwriting. What counts for JavaScript is that object will stay object

code/clue.js
```diff
@@ -1,59 +1,319 @@
...
 const randomSelector = array => {
   return array[Math.floor(Math.random() * array.length)]
 }
 
-// CREATE AN OBJECT THAT KEEPS THE MYSTERY.
-// With a killer, a weapon and a room.
-// The values will be set later.
+// Function to shuffle the suspects favorite weapon
+const shuffleFavoriteWeapon = () => {
+  favoriteWeapon = weapons[Math.floor(Math.random() * weapons.length)]; //Picks a random weapon from the array and creates a favoriteWeapon object
```
EmOlsz: You have the `randomSelector` function for that :)

code/clue.js
```diff
@@ -1,59 +1,319 @@
...
+  mystery.killer.favoriteWeapon = favoriteWeapon.name;// Assigns the name of favoriteWeapon to choosen killers object
+  document.getElementById('favoriteWeapon').innerHTML = `Prefered weapon: ${mystery.killer.favoriteWeapon}`; //Prints favoriteweapon to card
+}
 
-// This function will be invoked when you click on the killer card.
+// This function will be invoked in the loadingPickKiller function.
 const pickKiller = () => {
-  // This will randomly select a killer from the suspects. And add that to the mystery object.
-  mystery.killer = randomSelector(suspects)
+  document.getElementById("loader-wrapper-killer").style.display = "none";
```
EmOlsz, line 221: It's better to use class manipulation, not style manipulation

code/clue.js
```diff
@@ -1,59 +1,319 @@
...
+  document.getElementById('roomCard').style.backgroundColor = '#000'; 
+}
+// Calling the loadingPickRoom function when clicking the card
+document.getElementById("roomCard").onclick = loadingPickRoom;
+
+
+
+// Function to reveal mystery when clicking the button
+const revealMystery = () => {
+  if (mystery.killer === '') {
```
EmOlsz: Or we can simplify that : 

```
if (!mystery.killer) { ... }
```

For further reference check out https://www.freecodecamp.org/news/js-type-coercion-explained-27ba3d9a2839/

code/index.html
```diff
@@ -4,34 +4,67 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <link href="https://fonts.googleapis.com/css2?family=Creepster&family=Roboto:ital,wght@1,300&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="style.css" />
```
EmOlsz: `./` missing in path to file

code/index.html
```diff
@@ -4,34 +4,67 @@
...
+    </section>
 
-    <button>REVEAL THE CRIME</button>
+    <footer>
+      <p>Project 4, Cluedo Mystery Picker, by Anna Hellqvist @ Technigo Bootcamp 2020</p>
+      <p><a target="_blank" href="https://icons8.com/icons/set/magnifying-glass-tilted-left-emoji">Magnifying Glass Tilted Left icon</a> icon by <a target="_blank" href="https://icons8.com">Icons8</a></p>
+    </footer>
 
+  
     <script src="clue.js"></script>
```
EmOlsz: `./`

code/style.css
```diff
@@ -1,58 +1,176 @@
+/*** GENERAL STYLING ***/
+
 body {
   margin: 0;
-  font-family: "Courier New", Courier, monospace;
+  font-family: 'Roboto', sans-serif;
+  background: url(./assets/kamil-feczko-GhxWry42_zQ-unsplash.jpg);
```
EmOlsz: It's far more safe approach to wrap path in parentheses. Without them, some browsers might not find the path
`url("./assets/kamil-feczko-GhxWry42_zQ-unsplash.jpg");`

code/style.css
```diff
@@ -1,58 +1,176 @@
...
 }
 
-p {
-  margin: 1px;
-  font-weight: 300;
-  background-color: #fff;
+.magnifying-icon {
+  width: 48px;
+  height: 48px;
+  margin: 0px 20px;
```
EmOlsz: `margin: 0 20px;` is better, we don't have to write `px` if value is `0`

code/style.css
EmOlsz: Good catch Sara :)

code/style.css
```diff
@@ -1,58 +1,176 @@
...
-  background: crimson;
+  background: black;
+  color: white;
   border: none;
   font-size: 20px;
   border-radius: 8px;
   display: block;
   margin: 0 auto;
+  box-shadow: 2px 2px 2px white;
+  cursor: url(./assets/icons8-magnifying-glass-tilted-left-48.png),auto;
```
EmOlsz: `url(" ... ")`

EmOlsz (changes requested):
Good job Anna! I'm happy to see more than basic requirements :) Your code is pretty well structured.
There are some things you could improve, I've addressed them in comments. Please, take a look on them and provide changes. After that, I will be able to approve your project

code/style.css
AnnaHellqvist: Changed to class instead of id.

code/index.html
AnnaHellqvist: done! :)

code/clue.js
AnnaHellqvist: do you mean using querySelector or for ex use .className to add another class? Will the later override the previous values in the css? Or should I make two different classes for each? Not really sure how to best approach this :)

EmOlsz: Sorry, my comment might have been not clear enough :)
What I meant was that we should rather create class `hidden` like this :

```
.hidden {
    display: none;
}
```

and add it to your `loader-wrapper-killer` element in JS like that :

```
document.getElementById("loader-wrapper-killer").classList.add = "hidden";
```

Main idea behind that approach is that we should rather add or remove CSS classes instead of adding and removing exact styles

EmOlsz (approved):

code/clue.js
AnnaHellqvist: Ok thanks for the explanation, I will try this! 😄

PR #139, opened 2020-09-20
Cluedo - Destiny

destinyleia:


clue.js
```diff
@@ -0,0 +1,181 @@
...
+// YOU DON'T NEED TO CHANGE THIS, JUST TRY TO UNDERSTAND IT. AND HOW TO USE IT.
+const randomSelector = array => {
+  return array[Math.floor(Math.random() * array.length)]
+}
+
+// CREATE AN OBJECT THAT KEEPS THE MYSTERY.
+// With a killer, a weapon and a room.
+// The values will be set later.
+let crime = {
+ killer: [],
```
ClaireCaudwell: I think here you can also set each of the object properties to an empty string, null or undefined. Not sure if you can use [ ] as it may think you want to create an empty array? Maybe something the teachers can clarify? And not sure if you have to have them all as the same?

clue.js
```diff
@@ -0,0 +1,181 @@
...
+ weapon: [],
+ room: ''
+}
+
+// This function will be invoked when you click on the killer card.
+const pickKiller = () => {
+  // This will randomly select a killer from the suspects. And add that to the mystery object.
+  mystery.killer = randomSelector(suspects)
```
ClaireCaudwell, line 142: I think you need to replace "mystery" with "crime" as you've set the object with the empty properties as this name. So for each of these lines inside each of the functions it should be for example  `crime.killer = randomSelector(suspects)` or `document.getElementById('killerCard').style.background = crime.killer.color;`. Alternatively you could change the name of the crime object to mystery so you don't have to change all the lines of code inside each of the functions. By having the same name means you will be able to fill the empty crime object properties with this information for each card. Maybe it worked anyway?

clue.js
```diff
@@ -0,0 +1,181 @@
...
+
+  // This will change the background color of the card to the one connected to the chosen killer and show the full name of the killer. Feel free to add more things to show about the killer.
+  document.getElementById('killerCard').style.background = mystery.killer.color;
+  document.getElementById('killerCard').addEventListener('click', pickKiller)
```
ClaireCaudwell: Nice that you used the addEventListener here to call the function when clicked on, instead of writing it in the HMTL element. I think you might need to move this to outside the function so it can be called? Maybe the teacher can clarify this as I'm not sure.

clue.js
```diff
@@ -0,0 +1,181 @@
...
+
+
+// STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
+// 'The murder was committed by Jacob Green, in the living room with a rope.'
+
+function revealMystery() {
+document.getElementById("revealMystery").innerHTML = `The murder was committed by ${mystery.killer.firstName} ${mystery.killer.lastName}, in the ${mystery.room} with a ${mystery.weapon.name}`;
+}
+
+document.getElementById('killerCard').onclick = pickKiller;
```
ClaireCaudwell, line 177: I also called each function at the bottom of my code, so this is great, so they're all grouped together. But I think because you have the `document.getElementById('killerCard').addEventListener('click', pickKiller)` inside each of the functions, this would do this for you if you had them written outside of the function. Maybe something the teacher can confirm as I'm 100% sure!

clue.js
```diff
@@ -0,0 +1,181 @@
+// STEP 1 - CREATE OBJECTS FOR ALL THE SUSPECTS, SOMETHING LIKE THIS:
+
+const mrGreen = {
```
ClaireCaudwell, line 3: Your objects look nice and clearly defined with strings being used where necessary 💯

index.html
```diff
@@ -30,8 +31,9 @@
 
     <h1 id="mystery"></h1>
 
-    <button>REVEAL THE CRIME</button>
+    <button type="button" id="revealMystery" onclick="revealMystery()">REVEAL THE CRIME</button>
```
ClaireCaudwell: Here you've added the `onclick="revealMystery()"` which is great. This means you don't have to add this in the js file if you've already written it here. I'm not sure what the `type="button"` does or if you need it as you've already defined a button element? Maybe the teacher can confirm?

style.css
```diff
@@ -0,0 +1,89 @@
+body {
+  margin: 0;
+  font-family: "Courier New", Courier, monospace;
+  background-image: url('https://static.wixstatic.com/media/c88f3c_80f37c005f9a46209de3525512af4e1d~mv2.gif');
```
ClaireCaudwell, line 4: Nice tip you gave us about using a gif as a moving image! Looked great!

style.css
```diff
@@ -0,0 +1,89 @@
...
+  background-size: cover;
+  opacity: 0.9;
+  display: flex;
+  justify-content: space-around;
+  align-items: center;
+  border-radius: 15px;
+}
+
+.killer-deck {
+
```
ClaireCaudwell, line 29: Maybe good to remove this whitespace area so we can see easier that this is all together?

style.css
```diff
@@ -0,0 +1,89 @@
...
+  background: white;
+  border: none;
+  font-size: 20px;
+  border-radius: 8px;
+  display: block;
+  margin: 0 auto;
+  margin-bottom: 40px;
+}
+.button:hover,
+.button:focus {
```
ClaireCaudwell: Nice how you've written the same css class, but two different psuedo classes like this. Good tip for me!

style.css
```diff
@@ -0,0 +1,89 @@
...
+  box-shadow: inset 0 0 0 2em var(--hover);
+}
+
+.room-deck:hover {
```
ClaireCaudwell, line 76: Adding a box shadow when you hover over the card is a great idea 👍

index.html
```diff
@@ -14,13 +14,14 @@
         <p id="killerName"></p>
```
ClaireCaudwell, line 14: I see that the starting body tag is missing here, you have the end one below. Maybe is a good idea to pop that in?

ClaireCaudwell, line 14: Also the first article tag is missing for your killer card, but I can see you do have the end one. If you refer to the weapon deck you can see that it has it's start and end article tag. `<article class="weapon-deck" id="weaponCard"> <p>WEAPON</p>  <p id="weaponName"></p> <p id="weaponWeight"></p> </article>`

index.html
```diff
@@ -30,8 +31,9 @@
```
ClaireCaudwell, line 31: Here's you're also missing the end `</article>` tag for the room deck.

ClaireCaudwell (changes requested):
Hi Destiny! 

Nice to have the chance to look through your code. I think it looks really good, clear and concise. There are a few areas that I have left some comments for you to have a look at and see if you want to change yourself. Also I'm a bit unsure so it would be great if the teachers can give feedback regarding my comments to make sure that what I have written is actually right!

The only thing that I was missing was the readme information.  I found it really helpful to write the read me as a way of reflecting over the project, what I'd learnt and what I hadn't really understood.  It would be great to read your own reflections so I got an understanding of how you though it went etc. Also a link to your project, if you've deployed it on netlify, so I could have viewed it and then been able to understand why you chose to write the code in the way you did  :)

To update:
Readme including link to netlify. Once you've done this then you can save it as normal on your computer and then do the push request below which should push the changes to your technigo pull request. 

To update the PR:
Make code changes
git add -u
git commit -m "Fixed xxxxxxx"
git push/git push origin master
Your PR will be up-to-date

_redirects
```diff
@@ -0,0 +1 @@
+/* /index.html 200
```
EmOlsz: Is this file being used at all ??

clue.js
EmOlsz: Good assumption Claire. Yes, we shouldn't initialise those properties with empty array, because they are not going to be non-empty array later on. It's misrepresentation that could led to wrong assumption later on.

It's far more semantic approach to initialise mystery object with `null` values like this :

```
const mystery = {
  killer: null,
  room: null,
  weapon: null
}
```

Moreover, you can use `const` keyword - we will be changing values of properties, that's true, but JavaScript will not consider this as overwriting. What counts for JavaScript is that object will stay object

clue.js
EmOlsz: Definitely, it's good you noticed that Claire!

clue.js
EmOlsz: I checked whole file and you are right Claire - line 146 is 100% irrelevant and, being honest, it won't work. You cannot assign function to an event inside that funtion.
Still, you've done assignment in line 179, so everything work.
This line however (146) needs to be deleted, same goes for other functions in your code

index.html
EmOlsz: Good idea indeed!

index.html
```diff
@@ -3,7 +3,7 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <link rel="stylesheet" href="style.css" />
```
EmOlsz: `./` missing in path to file

index.html
EmOlsz: That's a lot of duplicating logic. You need to decide which approach you want to use Destiny

index.html
```diff
@@ -30,8 +31,9 @@
 
     <h1 id="mystery"></h1>
 
-    <button>REVEAL THE CRIME</button>
+    <button type="button" id="revealMystery" onclick="revealMystery()">REVEAL THE CRIME</button>
 
+    <footer></footer>
     <script src="clue.js"></script>
```
EmOlsz: `./`

style.css
```diff
@@ -0,0 +1,89 @@
+body {
+  margin: 0;
+  font-family: "Courier New", Courier, monospace;
```
EmOlsz, line 3: Good to see font fallback :)

style.css
```diff
@@ -0,0 +1,89 @@
...
+  border: none;
+  font-size: 20px;
+  border-radius: 8px;
+  display: block;
+  margin: 0 auto;
+  margin-bottom: 40px;
+}
+.button:hover,
+.button:focus {
+  box-shadow: inset 0 0 0 2em var(--hover);
```
EmOlsz: If you decide to copy code from internet, at least remember to read it first and analyse what will it do ;)
Over here you are using variable with name `hover`, but it's not defined anywhere

EmOlsz (changes requested):
I would like to say that in general it's good project Destiny, you've met requirements, congratulations :)
Nonetheless, there are some things you should fix right away. Please, take a look on both Claire's and my comments and provide changes. I would also like to mention that you should work a bit on quality of you commit messages - right now they are not meaningful.
After checking your fixes I will be able to approve project :)

EmOlsz:
Thank you Destiny, project approved :)

EmOlsz (approved):

PR #147, opened 2020-09-22
Tara-Cluedo project

trangtara:


code/index.html
```diff
@@ -4,33 +4,51 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <meta property="og:title" content="A game that turns you into a top-notch detective">
+    <meta property="og:description" content="Who is the killer? What is the weapon? and where is the real location of the murder?">
+    <meta property="og:image" content="https://detectivegame.netlify.app/assets/sherlock.jpg">
+    <meta property="og:url" content="https://detectivegame.netlify.app/">
```
thereselinden, line 10: Wow, I really like this feature

code/index.html
```diff
@@ -4,33 +4,51 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Ranchers&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="style.css" />
```
thereselinden, line 12: Don't forget to add a ./ when you are referring to your css file. Such as "./style.css

code/index.html
```diff
@@ -4,33 +4,51 @@
...
+        </article>
+      </section>
+      <div class="reveal">
+        <img id="detectiveImg" class="detectiveImage" src="./assets/detective 2.png" height="150">
+        <h1 id="mystery" class="revealMessage"></h1>
+      </div>
+      <button id="button" onclick="revealMystery()">REVEAL THE CRIME</button>
+    </div>
 
     <script src="clue.js"></script>
```
thereselinden, line 53: Great that you have linked the script at the end, so your content can load before your script.  Don't forget to add ./ to indicate that the file is in the same folder. (same as for css file)

code/index.html
```diff
@@ -4,33 +4,51 @@
...
-        <p id="roomName"></p>
-      </article>
-    </section>
-
-    <h1 id="mystery"></h1>
-
-    <button>REVEAL THE CRIME</button>
+    <div class="wrap">
+      <header class="header">
+        <img class=logo src="./assets/cluedoLogo.png" height="100">
```
thereselinden, line 18: Here you are using ./ to indicate the same folder, great! The height attribute is inline css and should be moved to CSS-file

code/index.html
```diff
@@ -4,33 +4,51 @@
...
+          <p class="cardContent"id="killerName"></p>
+          <p class="cardContent"id="killerAge"></p>
+          <p class="cardContent"id="killerOccupation"></p>
+          <img id="killerImage" src="" height="200" />
```
thereselinden, line 29: Same here regarding the height attribute, move it to css-file

code/index.html
```diff
@@ -4,33 +4,51 @@
...
+          <svg class="killerIcon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512"><path d="M376.61 54.46c-82.95-72.61-222.26-72.61-305.22 0C7.36 110.5-31.01 224.44 32.63 416 64.53 512 224 512 224 512s159.47 0 191.37-96c63.64-191.56 25.27-305.5-38.76-361.54zM288 64c8.84 0 16 7.16 16 16s-7.16 16-16 16-16-7.16-16-16 7.16-16 16-16zm-128 0c8.84 0 16 7.16 16 16s-7.16 16-16 16-16-7.16-16-16 7.16-16 16-16zm16 400c-8.84 0-16-7.16-16-16s7.16-16 16-16 16 7.16 16 16-7.16 16-16 16zm0-64c-8.84 0-16-7.16-16-16s7.16-16 16-16 16 7.16 16 16-7.16 16-16 16zm0-64c-8.84 0-16-7.16-16-16s7.16-16 16-16 16 7.16 16 16-7.16 16-16 16zm-48-64c-35.35 0-64-28.65-64-64 0-17.67 14.33-32 32-32h64c17.67 0 32 14.33 32 32 0 35.35-28.65 64-64 64zm96-128c-8.84 0-16-7.16-16-16s7.16-16 16-16 16 7.16 16 16-7.16 16-16 16zm48 320c-8.84 0-16-7.16-16-16s7.16-16 16-16 16 7.16 16 16-7.16 16-16 16zm0-64c-8.84 0-16-7.16-16-16s7.16-16 16-16 16 7.16 16 16-7.16 16-16 16zm0-64c-8.84 0-16-7.16-16-16s7.16-16 16-16 16 7.16 16 16-7.16 16-16 16zm48-64c-35.35 0-64-28.65-64-64 0-17.67 14.33-32 32-32h64c17.67 0 32 14.33 32 32 0 35.35-28.65 64-64 64z"/></svg>
+          <p class="cardContent"id="killerName"></p>
```
thereselinden: Good with the naming of id:s where you are using camelCase, don't forget to use kebab-case when naming your classes.

code/index.html
```diff
@@ -4,33 +4,51 @@
...
-        <p>ROOM</p>
-        <p id="roomName"></p>
-      </article>
-    </section>
-
-    <h1 id="mystery"></h1>
-
-    <button>REVEAL THE CRIME</button>
+    <div class="wrap">
+      <header class="header">
```
thereselinden, line 17: I think your code is easy to read and follow. You have a good habit with indentation, great job =)

code/style.css
```diff
@@ -1,58 +1,233 @@
+* {
+  box-sizing: border-box;
+}
+
 body {
```
thereselinden, line 5: Here you could have added the background color of #c44536 so you don't have to write the same line of code several times. (works now since you are having the same background color on your page.)

code/style.css
```diff
@@ -1,58 +1,233 @@
...
+.cardContent {
   margin: 1px;
   font-weight: 300;
   background-color: #fff;
 }
 
+.killerIcon {
+  margin-top: 60px;
+  width: 50%;
+  fill: rgba(255,255,255,0.5);
```
thereselinden, line 40: Really like that you are using rgba! Looks great!

code/style.css
```diff
@@ -1,58 +1,233 @@
...
+  grid-template-columns: 1fr 1fr 1fr;
   align-items: center;
+  justify-items: center;
 }
 
-.killer-deck {
-  width: 250px;
-  height: 400px;
-  background: #000;
+.deck {
```
thereselinden, line 76: I really like that you have styled all cards in one. =)

It would be great with a comment here to tell that you are merging the three different cards. Or, if you would have used kebab-case when naming the classes you could have written this style as: 
.deck-killer-card, .deck-weapon-card, .deck-room-card {
}

code/style.css
```diff
@@ -1,58 +1,233 @@
...
+  align-items: center;
+  border-radius: 16px;
+}
+
+.detectiveImage {
+  width: auto;
+  margin-left: 30px;
 }
+
+.shakeX {
```
thereselinden, line 136: Really like your animation, great job! =)

code/style.css
```diff
@@ -1,58 +1,233 @@
...
+
 .cards {
-  margin: 50px auto;
-  width: 90%;
-  height: 600px;
-  background: #ccc;
-  display: flex;
-  justify-content: space-around;
+  margin: 30px;
+  display: grid;
```
thereselinden, line 70: Like that you are trying grid instead of flexbox here! =)

code/clue.js
```diff
@@ -6,54 +6,271 @@ const mrGreen = {
...
-// STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
-// 'The murder was committed by Jacob Green, in the living room with a rope.'
+    document.getElementById('roomName').innerHTML = `${mystery.room}`
+  }, 1500);
+}
+
+const revealMystery = () => {
+  document.getElementById('mystery').innerHTML = `The murder was committed by ${mystery.killer.firstName} ${mystery.killer.lastName}, a ${mystery.killer.age}-year-old ${mystery.killer.occupation}, in the ${mystery.room} with a ${mystery.weapon.name} of ${mystery.weapon.weight} pounds.`;
+  document.querySelector('.reveal').classList.add('active');
+}
```
thereselinden: It's very easy to follow your code here too. Logical order in your functions. Great job! 
One think to add in other projects is to have a little description what the function actually does, then it would make even more sense =)

code/clue.js
```diff
@@ -6,54 +6,271 @@ const mrGreen = {
...
+      favouriteWeapon,
+    ]
+    mystery.weapon = randomSelector(weaponsClone);
+  
+    document.getElementById('weaponName').innerHTML = `${mystery.weapon.name} of ${mystery.weapon.weight} pounds`
+  }, 1500);
 }
 
-// CREATE FUNCTIONS pickWeapon and pickRoom in a similar way.
+const pickRoom = () => {
```
thereselinden, line 255: A good thing to add on your pickRoom, pickWeapon, pickKiller function is to force them to be open in the right order, otherwise your logic will be broken if I open weapon card before I open the killer card and so on. 

One solution could be to give them a false value and return the value as true once the card is clicked. 
Example is to have variables at top of the script and give them a boolean value of false. 
killerPicked = false
weaponPicked = false
roomPicked = false

once the code is invoked the value is changed to true "killerPicked = true" (which is written at the bottom of the pickKiller function.

code/clue.js
```diff
@@ -6,54 +6,271 @@ const mrGreen = {
...
+    ]
+    mystery.room = randomSelector(roomsClone);
 
-// STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
-// 'The murder was committed by Jacob Green, in the living room with a rope.'
+    document.getElementById('roomName').innerHTML = `${mystery.room}`
+  }, 1500);
+}
+
+const revealMystery = () => {
```
thereselinden, line 285: Since I'm allowed to click the reveal crime button before I have clicked the cards, the return will be undefined. If you add an if / else statement here then you can return "no mystery has been solved" if no cards has been selected. And if they have been selected, you print out by who, whit what and in which room.

thereselinden:
Hi Tara! 
It was a pleasure to read your code. . 
Overall you have great logic in your code and its easy to read. 
I like your style and especially your animation effect on the cards! 

I would suggest that you force the cards to be open in correct order so that your logic can be used properly. Also to not make the reveal button print out a string of undefined data when cards is not selected.

thereselinden (approved):

code/clue.js
```diff
@@ -6,54 +6,271 @@ const mrGreen = {
...
+  'Observatory', 
+  'Theater', 
+  'Guest House', 
+  'Patio',
+]
 
-// THIS FUNCTION WILL RANDOMLY SELECT ONE ITEM FROM THE ARRAY THAT YOU PASS IN TO THE FUNCTION.
-// YOU DON'T NEED TO CHANGE THIS, JUST TRY TO UNDERSTAND IT. AND HOW TO USE IT.
-const randomSelector = array => {
+const randomSelector = (array) => {
```
EmOlsz: Since you pass only one argument, you don't need parentheses around it

```
const randomSelector = array => { ... }
```

code/clue.js
```diff
@@ -6,54 +6,271 @@ const mrGreen = {
...
   return array[Math.floor(Math.random() * array.length)]
 }
 
-// CREATE AN OBJECT THAT KEEPS THE MYSTERY.
-// With a killer, a weapon and a room.
-// The values will be set later.
+const mystery = {
+  killer: '',
+  weapon: '',
```
EmOlsz: It's even more semantic approach to initialise mystery object with `null` values like this :

```
const mystery = {
  killer: null,
  room: null,
  weapon: null
}
```

Moreover, it's great you used `const` keyword - we will be changing values of properties, that's true, but JavaScript will not consider this as overwriting. What counts for JavaScript is that object will stay object

code/clue.js
```diff
@@ -6,54 +6,271 @@ const mrGreen = {
...
 
-  // This will change the background color of the card to the one connected to the chosen killer and show the full name of the killer. Feel free to add more things to show about the killer.
-  document.getElementById('killerCard').style.background = mystery.killer.color
-  document.getElementById(
-    'killerName'
-  ).innerHTML = `${mystery.killer.firstName} ${mystery.killer.lastName}`
+  setTimeout(function() {
+    document.querySelector('.killer-deck').classList.remove('shakeX');
+    document.querySelector('.killer-deck').classList.remove('bounce');
+    document.querySelector('.killer-icon').classList.add('hide');
```
EmOlsz: a lot of reference to the same HTML element - this is great use case for variable, don't you think?

code/clue.js
```diff
@@ -6,54 +6,271 @@ const mrGreen = {
...
+
+    shuffleFavouriteWeapon();
+    mystery.killer = randomSelector(suspects)
```
EmOlsz: You seem to use semicolons only from time to time - we have to be consequent about that

code/index.html
EmOlsz: exactly :)

code/style.css
```diff
@@ -1,58 +1,233 @@
...
+  justify-content: space-around;
+  align-items: center;
+}
+
+.slogan {
+  text-align: center;
+  font-size: 45px;
+  padding: 10px;
+  color: white;
+  font-family: 'Ranchers', cursive;
```
EmOlsz, line 27: Good to see font fallback :)

code/style.css
```diff
@@ -1,58 +1,233 @@
...
   background-color: #fff;
 }
 
+.killer-icon {
+  margin-top: 60px;
+  width: 50%;
+  fill: rgba(255,255,255,0.5);
+}
+
+.killer-icon.hide {
```
EmOlsz: defining class `.hide` would be enough, you don't need `.killer-icon` for that

code/style.css
```diff
@@ -1,58 +1,233 @@
...
-  background: #000;
+.deck {
+  max-width: 250px;
+  height: 433px;
+  background: #ffd046;
   text-align: center;
   padding-top: 30px;
   border-radius: 16px;
+  border: solid 2px black;
+  box-shadow: 10px 10px 5px 0px rgba(0, 0, 0, 0.3)
```
EmOlsz: We don't need to use `px` if value is `0` :

```
  box-shadow: 10px 10px 5px 0 rgba(0, 0, 0, 0.3);
```

Moreover, you forgot about semicolon

code/style.css
```diff
@@ -1,58 +1,233 @@
...
+  width: auto;
+  margin-left: 30px;
 }
+
+.shakeX {
+  animation-name: shakeX;
+  animation-duration: 1s;
+}
+
+@keyframes shakeX {
```
EmOlsz, line 141: Great job Tara :)

EmOlsz (changes requested):
Good job Tara, congrats :) You made proper project this week. Besides JS, I also like that you experiment with svg's and CSS in general!
Before I'll be able to approve your project, you will need to provide some changes - take a look on my comments and add fixes!

EmOlsz (approved):

PR #148, opened 2021-01-03
Cluedo W.5 - fixed 

ckjerrgren:
Updated code according to maks instructions.


code/clue.js
```diff
@@ -4,56 +4,191 @@ const mrGreen = {
...
+};
+document.getElementById("weaponCard").onclick = pickWeapon;
+
 
 // STEP 4 - CREATE A FUNCTION revealMystery that will be invoked when you click that button. It should show something like:
 // 'The murder was committed by Jacob Green, in the living room with a rope.'
+
+
+
+function revealMystery () {
```
EmOlsz: This functionality is still not fixed - the difference is that when we click "Reveal the crime" button now, we got error in Console tab.
I know you have far more important tasks to focus on right now Claudia, so there is a tip for you how to solve this issue : you need to prevent function from being called until all cards (killer, weapon, room) are selected. How can we do that? We know that when all cards are selected, both `mystery.killer`, `mystery.weapon` and `mystery.room` are not equal to `null`.

EmOlsz (changes requested):

ckjerrgren:
Now its updated with an if/ else statement for revealing the mystery

HIPPIEKICK (approved):

